PR #7, opened 2022-06-08
fix: expose ability to set logger level

stevearc:
## Summary
The current logger is unable to change the logging level

The class structure did not seem to be set up correctly. Tracing the logic from the existing `require('neotest.logging'):set_level('WARN')`:

1. `require('neotest.logging')` resolves to the result of `Logger:new()`
2. `Logger:new()` effectively returns `setmetatable(logger, Logger)`
3. There is no `__index` field on the Logger metatable, so looking up `set_level` will return nil

The fix is to `setmetatable(logger, { __index = Logger })`, and to also set the `_level` onto the logger object instead of the Logger class. It kind of doesn't matter because they're both effectively singletons, but it's technically more correct? ¯\\_(ツ)_/¯

I also slipped in a call to set the default log location to `stdpath('log')` if available (upcoming in nightly, see https://github.com/neovim/neovim/pull/15583).

Would potentially be nice to set this from a config option, but I think that's outside the scope of this PR.

## Test plan
In init.lua added:
```lua
require("neotest.logging"):set_level("WARN")
require("neotest").setup({})
```
Ran tests
Observed no output in ~/.cache/nvim/neotest.log

Tested again without the `set_level` line, and confirmed that the debug logs appear in the logfile.

### DIFF
--- a/lua/neotest/logging.lua
+++ b/lua/neotest/logging.lua
@@ -23,28 +23,32 @@ local log_date_format = "%FT%H:%M:%SZ%z"
 local Logger = {}
 
 ---@return neotest.Logger
-function Logger:new(filename, opts)
+function Logger.new(filename, opts)
   opts = opts or {}
   local logger = loggers[filename]
   if logger then
     return logger
   end
   logger = {}
-  setmetatable(logger, self)
+  setmetatable(logger, { __index = Logger })
   loggers[filename] = logger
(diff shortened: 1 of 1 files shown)


rcarriga:
Thanks for the PR! Logging was completely thrown in without much thought, need to tidy it up a bit anyway

PR #14, opened 2022-06-09
fix: flaky treesitter parsing of tests

stevearc:
## Motivation
The symptom I was seeing was if I opened a plenary test file:
```lua
describe("Busted", function()
    it("should succeed test", function()
      assert.truthy("Yup.")
    end)
end)
```
Then opened the summary
single test

Then duplicated the tests and renamed them so the file looks like this:
```lua
describe("Busted", function()
    it("should succeed test", function()
      assert.truthy("Yup.")
    end)
end)

describe("Busted too", function()
    it("should succeed test", function()
      assert.truthy("Yup.")
    end)
end)
```
Then `:w` to update the tests, the neotest summary would *clear* all tests and the file would appear as having no tests.
sad test

I added a bunch of logging, and AFAICT the buffer contents, query, parser, absolutely everything all looks good and identical to when the parser succeeds. The only difference is that the query fails to match any nodes.

## Summary
I noticed that we were fetching the file content, setting it into a scratch buffer, parsing, then discarding the scratch buffer. This seems like it's trying to optimize for the case where vim doesn't have (and doesn't want) the file to be opened as a buffer. I looked through `client/init.lua`, and I can only find cases where we're calling `adapter.discover_positions` on directories or files that are already opened (triggered from `BufAdd`, `BufWritePost`).

The fix replaces the scratch buffer logic with a call to `bufadd()` and `bufload()`, so we can use existing buffers when already loaded. While I don't understand the root cause of the problem, this change seems to fix it.

If we do need to preserve the "clean up buffers once we're done parsing them" logic, we can still do that by unloading the buffer once the parsing is completed (IFF it was unloaded before parsing).

## Test plan
Followed the repro steps in [Motivation](#motivation) and saw that the tests appear as expected
happy test

### DIFF
--- a/lua/neotest/lib/treesitter/init.lua
+++ b/lua/neotest/lib/treesitter/init.lua
@@ -115,13 +115,13 @@ end
 ---@param buf integer
 ---@return neotest.Tree
 local function parse_buf_positions(file_path, query, buf, opts)
-  local ft = require("neotest.lib").files.detect_filetype(file_path)
+  local ft = vim.api.nvim_buf_get_option(buf, "filetype")
   local ts_parsers = require("nvim-treesitter.parsers")
   local lang = ts_parsers.ft_to_lang(ft)
   local parser = ts_parsers.get_parser(buf, lang)
   local root = parser:parse()[1]:root()
   if type(query) == "string" then
-    query = vim.treesitter.parse_query(ft, query)
+    query = vim.treesitter.parse_query(lang, query)
   end
   local positions = collect(file_path, query, buf, root)
   local structure = parse_tree(positions, {}, opts)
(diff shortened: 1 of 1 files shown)


rcarriga:
I can't replicate the original issue given the steps above unfortunately. It's very odd that you're seeing this and that it's fixed by this change.

> This seems like it's trying to optimize for the case where vim doesn't have (and doesn't want) the file to be opened as a buffer.

This is also to avoid having buffers that are modified but not written (which of course test runners won't see)

> I looked through client/init.lua, and I can only find cases where we're calling adapter.discover_positions on directories or files that are already opened (triggered from BufAdd, BufWritePost).

Actually when a directory is found, all files within it are read as well (see https://github.com/rcarriga/neotest/blob/master/lua/neotest/client/init.lua#L423-L438), so we definitely do need the clean up.

I'd rather figure out the root cause of the issue than apply a blind fix. Unfortunately there's not much I can do though until I can replicate, so if you could point to a repo that reproduces that would be great.

stevearc:
> I can't replicate the original issue given the steps above unfortunately. It's very odd that you're seeing this and that it's fixed by this change.

Agree on both counts. Since you can't repro, I'll cut down my neovim to a minimal config and try harder to find the root cause.

> This is also to avoid having buffers that are modified but not written (which of course test runners won't see)

Ah interesting. I think that's a valid approach, but I would also raise the perspective that since neotest is integrated into the editor it makes sense for it to reflect the state of the editor rather than the state of files on disk. No matter what, running tests in an unsaved file is going to produce unexpected results. I could see it going either way, though.F

> Actually when a directory is found, all files within it are read as well (see https://github.com/rcarriga/neotest/blob/master/lua/neotest/client/init.lua#L423-L438), so we definitely do need the clean up.

D'oh! Thanks for pointing that out. I'm still learning the codebase 😅

Will dig deeper and report back

stevearc:
Well it's been a rough debugging session, but I've found two key details.
1. The config causing this issue is the treesitter highlight module. Enable it, and I get the repro. Disable it, and it works fine.
2. The reason there are no matches in the failure case is because treesitter parsing fails (produces an ERROR node as the root).

I have no idea why the parsing would fail, and certainly no idea why toggling the highlight module would have any impact on it. I think I've reached the end of the road on this one 😅

It may not have all been useless though! In the midst of throwing everything at the problem I found a pretty significant optimization. I think it makes more sense to make a new PR rather than rebranding this one, so I'll move it over to #17

PR #16, opened 2022-06-09
docs(adapter): add info about args

stevanmilic:


rcarriga:
Appreciate the PR but this actually just a feature of neotest-python :sweat_smile: There is no standard interface for configuring an adapter

PR #17, opened 2022-06-10
perf: use treesitter string parser for test discovery

stevearc:
I stumbled across the approach while fiddling around with the treesitter parsing logic. Currently we're reading a file, splitting it by newline, inserting it into a new buffer, parsing the buffer, then deleting the buffer. If we use a TS string parser instead, we just read the file and parse the contents directly. My very unscientific benchmark of opening the summary in the cpython repo and waiting for results to populate took the search time down from ~10s to ~2s. Seems like a win!

### DIFF
--- a/lua/neotest/lib/treesitter/init.lua
+++ b/lua/neotest/lib/treesitter/init.lua
@@ -20,10 +20,10 @@ end
 
 ---@param file_path string
 ---@param query table
----@param buf integer
+---@param source string
 ---@param root table
 ---@return neotest.FIFOQueue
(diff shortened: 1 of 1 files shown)


rcarriga:
Wow this is fantastic, I remember writing this wishing I could parse the string directly... Now I feel a bit silly :sweat_smile: Thank you for this!

rcarriga:
Does this have effect on the issue you were having in #14?

stevearc:
Unfortunately it did not 😅

Issue #13, opened 2022-06-09
Performance problems on large repos


stevearc:
When I try to open the summary on a test file in a repo at work, nvim freezes for a very long time.
The issue is this line:
https://github.com/rcarriga/neotest/blob/aaf2107a0d639935032d60c49a5bdda2de26b4d6/lua/neotest/client/init.lua#L408
The find_files actually completes within ~2 seconds, but then running `adapter.is_test_file` 250k+ times takes several minutes (vim-test adapter). After that freeze, there is a second freeze while it tries to write all the events to the logfile (hasn't finished yet).

My personal opinion, test discovery should be definitely optional, and possibly configurable. My previous job had a monorepo that was so big it was only served as a virtual filesystem, so running any type of crawling operation would have terrible side effects. Making it configurable (e.g. only these dirs, only this depth) might be nice, but would probably make more sense per-adapter instead of globally. If adapters need to control the search path, their API would have to change from testing individual files that neotest gives them to calling back into neotest with whatever their search config is. IDK if that refactor is worth it, which is why I could go either way on making the search configurable.

Another direction could be to customize the search on a per-directory basis instead of per-adapter. That avoids the need for an adapter API refactor, and in general this kind of functionality would be incredibly useful. I often work with many different types of repos on the same machine, sometimes in the same vim instance (using `:tcd` and one tab per project), and these projects will sometimes use the same language but require different configurations to run tests. I'd love to be able to configure my test adapters on a per-project basis. A rough proposal, it could look like:
```lua
require('neotest').setup({
  adapters = ...,
  summary = ...,
  discovery = {
    enabled = true,
  },
  projects = {
    ["~/work/big"] = {
      adapters = ...,
      discovery = {
        enabled = true,
        dirs = {'tests/unit', 'frontend/tests'},
        depth = 2
      },
    },
    ["~/personal/project"] = {
      adapters = ...,
      discovery = {
        enabled = false,
      },
    },
  },
})
```

I am happy to submit a PR for any parts of this once we align on a solution

Unrelated question: I'll probably be making more proposals, requests, and questions. Is filing an issue the best way to start a discussion, or would you prefer some other channel?

stevearc:
> the adapters I've written (and I imagine others) would need to be refactored to handle configuring multiple instances

[...]

PR #30, opened 2022-06-15
Allow disabling directory crawling for tests

stevearc:
Intended to address #13

Relatively straightforward: adds a config option that, when disabled, doesn't call `_update_positions` on directories. Files still get populated in the summary on `BufOpen`/`BufWritePost`, and I also added some logic to scan the files that vim opens on start.

## Test plan

New functionality
- Set `discovery.enabled = false`
- Open a test file in a directory with multiple test files
- Observe that the test summary only displays the currently-open file

Nothing broke
- Set `discovery.enabled = true`
- Open a test file in a directory with multiple test files
- Observe that the test summary displays all discoverable tests, as per usual

## Future work
Per-directory configuration would still be a great addition. Another feature that could be cool is some test adapters are capable of running tests on a directory. It would be great if we could trigger a run on an entire directory without scanning it first, and then populate the summary based on the results. I may start looking into that next.

### DIFF
--- a/lua/neotest/client/init.lua
+++ b/lua/neotest/client/init.lua
@@ -537,7 +537,9 @@ function NeotestClient:_start()
         if not adapter_id then
           return
         end
-        self:_update_positions(lib.files.parent(file_path), { adapter = adapter_id })
+        if config.discovery.enabled then
+          self:_update_positions(lib.files.parent(file_path), { adapter = adapter_id })
+        end
       end
       self:_update_positions(file_path, { adapter = adapter_id })
     end)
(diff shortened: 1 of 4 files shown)


rcarriga:
Thanks for putting this together! I'll probably have time over the weekend to review properly :+1:

stevearc:
Figured out the issue with tests not updating. Will address in a separate PR

stevearc:
Oh I didn't mention this directly, but this PR is ready for review. It had some unfortunate side effects before the position merging, but now it works fine.

lua/neotest/client/init.lua
```diff
@@ -568,6 +572,14 @@ function NeotestClient:_start()
   end)
 
   self:_update_adapters(async.fn.getcwd())
+  -- If discovery is not enabled, we need to update positions for all open
+  -- buffers on startup
+  if not config.discovery.enabled then
+    for _, bufnr in ipairs(async.api.nvim_list_bufs()) do
```
rcarriga, line 578: Let's add a test for this case.

rcarriga (changes requested):
This is very nice, and works brilliantly with the new merging code. Thanks for putting this together!

Just want to add the one test case and then LGTM

stevearc:
Done!

PR #32, opened 2022-06-17
Enable position merging for more cases

stevearc:
While debugging #30, I discovered that the reason tests weren't updating on save was because the root tree was a `type = 'file'`, and merging `type = 'file'` _into_ `type = 'file'` did not do anything.

This PR adds the following capabilities:
* Merging file into file works as expected
* Merging a _new_ file into a dir works (previously produced error)
* Merging two dirs with no common root joins them both under a new tree based at the common ancestor

I've added tests for everything, but let me know if there are other cases that you think should be covered.

### DIFF
--- a/tests/unit/lib/positions/init_spec.lua
+++ b/tests/unit/lib/positions/init_spec.lua
@@ -244,4 +244,37 @@ describe("merge", function()
     }
     assert.are.same(expected, ret:to_list())
   end)
+
+  it("merge(dir, file) adds new file", function()
+    local dir = create_tree({
+      { type = "dir", id = "/root", path = "/root" },
+    })
+    local file = create_tree({
+      { type = "file", id = "/root/file", path = "/root/file" },
+      { type = "test", id = "test-2", path = "/root/file" },
+    })
+    local ret = positions.merge(dir, file)
+    local expected = {
+      {
+        id = "/root",
+        path = "/root",
+        type = "dir",
+      },
+      {
+        {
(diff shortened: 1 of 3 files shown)


stevearc:
Ugh...unfortunately I have found another merge issue but I'm at the end of my time for tonight. Will come back to this when I get a chance.

stevearc:
Found the problem. Should be ready to review now! 🤞

tests/unit/lib/positions/init_spec.lua
```diff
@@ -0,0 +1,344 @@
...
+      },
+      {
+        { type = "dir", id = "/root/sub2", path = "/root/sub2" },
+        {
+          { type = "file", id = "/root/sub2/file", path = "/root/sub2/file" },
+          { { type = "test", id = "test-2", path = "/root/sub2/file" } },
+        },
+      },
+    }
+    print(string.format("ret: %s", vim.inspect(ret:to_list())))
```
rcarriga, line 341: Rogue print :sweat_smile:

rcarriga (changes requested):
This is fantastic! The tests are also very much appreciated. The tree object is a bit nasty so it's nice to have some safety around it.

Just the one line to remove then LGTM!

tests/unit/lib/positions/init_spec.lua
stevearc: Dangit! I need to add a precommit hook for these 😅

PR #49, opened 2022-07-01
Validate window before closing to avoid invalid window id errors

wookayin:
The window that is going to be closed (via `win:close()` call)
probably have already been closed or no longer available, in which case
"Invalid window id" errors will occur when closing the window.

To avoid such an error, we can if the window is valid right before
closing the window.

### DIFF
--- a/lua/neotest/lib/ui/float.lua
+++ b/lua/neotest/lib/ui/float.lua
@@ -76,6 +76,9 @@ function Float:close(force)
   if not force and api.nvim_get_current_win() == self.win_id then
     return false
   end
+  if not vim.api.nvim_win_is_valid(self.win_id) then
+    return true
+  end
   local buf = self:get_buf()
   pcall(api.nvim_win_close, self.win_id, true)
   for _, listener in pairs(self.listeners.close) do


wookayin:
FYI, here is some scenario I've run into:

- `lua require("neotest").output.open()` to open a floating window
- `:vsplit` on the main editor  (due to an autocmd, the floating window will be automatically hidden)
- The "invalid window id" error happens

wookayin:
If the return value of `win:close` is `false`, it will create a tons of WinEnter autocmds, so fa6cfe5..6b8e319 has changed and corrected the return value.

rcarriga:
Thanks for the PR! Could you run `scripts/style` and remove the comment, it's obvious enough what the check does.

The failing tests are unrelated.

wookayin:
@rcarriga I've updated the PR as requested.

rcarriga:
Sorry I've just noticed that the close win call is already wrapped in `pcall` which means an error shouldn't be raised. Are you sure this is where the error came from? I've tried reproducing and couldn't

wookayin:
Hmm yeah I now see the pcall when calling nvim_win_close API. It's indeed strange. I should have saved the stacktrace when running into the bug, but what I remember was the stacktrace certainly said `neotest` and the error was happening when the output (floating) window was being auto-closed, so I assumed this was inside `win:close()`.

As you might already know such invalid window error isn't always reproducible in a deterministic fashion, as often involving race conditions and is intertwined with other plugins. I cannot reproduce it at the moment. I will try using the unpatched version and get back again when I run into the same error.

wookayin:
Let me close this for now, but I will reopen this when I identify the root cause again. Thanks.

PR #54, opened 2022-07-04
Update README with link to testthat adapter

shunsambongi:
Hi, thanks for this great plugin! I wrote an adapter for the testthat package for R.


rcarriga:
Thanks for contributing! :smile:

PR #69, opened 2022-07-16
feat(ui): add NeotestUnknown highlight group

kyazdani42:


lua/neotest/config/init.lua
```diff
@@ -116,6 +116,7 @@ local default_config = {
     select_win = "NeotestWinSelect",
     marked = "NeotestMarked",
     target = "NeotestTarget",
+    unknown = "NeotestUnknown",
```
rcarriga, line 119: Add to the definitions above as well

rcarriga (changes requested):
Thanks for the PR! Just the one extra change and then LGTM

rcarriga:
Thanks again!

PR #77, opened 2022-07-25
feat: add config option to specify default strategy

stevearc:
I've written [a custom neotest strategy](https://github.com/stevearc/overseer.nvim/blob/master/lua/neotest/client/strategies/overseer.lua) and I'd like to use it by default. I tested both by setting the `default_strategy` and leaving it empty in the call to `setup()`, and both cases work as expected.

### DIFF
--- a/lua/neotest/config/init.lua
+++ b/lua/neotest/config/init.lua
@@ -26,7 +26,7 @@ vim.cmd([[
 ---@field icons table<string, string>
 ---@field highlights table<string, string>
 ---@field floating neotest.Config.floating
----@field default_strategy string
+---@field default_strategy string | function
 ---@field strategies neotest.Config.strategies
 ---@field summary neotest.Config.summary
 ---@field output neotest.Config.output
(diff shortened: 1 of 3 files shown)


lua/neotest/config/init.lua
```diff
@@ -26,6 +26,7 @@ vim.cmd([[
 ---@field icons table<string, string>
 ---@field highlights table<string, string>
 ---@field floating neotest.Config.floating
+---@field default_strategy string
```
rcarriga, line 29: This can also be a function that returns a `neotest.Process` as well. I haven't added the process type to the doc file yet so I'd be happy just saying it can be `string | function` and let users discover themselves since they'll need to be adventurous to implement :sweat_smile:

rcarriga (changes requested):
Very cool :grin: Hadn't actually thought of that as a way to have external strategies!

Just the one type fix and then LGTM

stevearc:
Done!

Issue #76, opened 2022-07-25
Re-run tests from custom strategy


stevearc:
Hi again! I recently polished up my [task runner plugin](https://github.com/stevearc/overseer.nvim) enough for general release, and I've written a custom [neotest strategy](https://github.com/stevearc/overseer.nvim/blob/master/lua/neotest/client/strategies/overseer.lua) for it that allows neotest tests to run within a task. This is already working for simple cases (including for the streaming results!) 

The one thing I haven't gotten working yet is the ability to restart a test run from a task. There are some affordances for "restart task", "restart task on buffer save", "restart task on failure", etc. and at the moment they re-run the test, but neotest doesn't get the new results. The issue is that the strategy results are _pull based_, where the runner calls into it and expects to receive results. I have two ideas for how to rework this, and I'd love to get your opinion on if either of them would be an appropriate change, or if you have a better idea.

1. **Pass the tree and args into the strategy**. Since the tree and args are all that's needed by the client to kick off a new run, giving them to the strategy would allow it to repeat the same run. On the strategy side, I could then do some magic to detect when a task should be re-used instead of creating a new one. This would be a pretty minimal change for neotest, but one thing I'm not sure of is if the tree parameter can become "stale" after mutations.
2. **Refactor neotest to make strategy results push-based**. The runner would kick off the strategy, and then the strategy would be responsible for calling back into neotest to notify the client of new results. This would require more extensive changes within neotest and a refactoring of existing strategies.

rcarriga:
Hmm right now the strategies are very unaware of most of neotest, they just run commands and provide raw output. Having them be responsible for starting off tests etc would be drastic change of responsibility which I don't think would make sense.

It seems more like the better integration point for you would be a custom consumer.  Your consumer could have the same interface as the `run` consumer but then also contain the logic required for rerunning tests. We could expose how the run consumer parses the args to avoid duplication.

stevearc:
My gut reaction was that it doesn't make much sense for overseer to be a consumer, but after thinking about it more I think it could work if I kept it as the strategy, but _also_ made a consumer like the `run` consumer. I'll play with this idea a bit and see if I can get it doing what I want. Thanks for the suggestion!

stevearc:
Got it working! I had to do a little gross monkey patching of `neotest.run` to make sure it gets used when you run tests from the summary, but that's the only bit I'm unhappy about.

[...]

PR #83, opened 2022-08-02
refactor: expose neotest.run.get_tree_from_args

stevearc:
Following up on #76, it turns out I actually _would_ really like this helper function to be exposed.


rcarriga:
Sounds good :+1:

PR #84, opened 2022-08-02
fix: crash in diagnostics consumer

stevearc:
Small bug introduced in 4fdce02df7046b7f8be8b112377d8c65593f2553 that broke the diagnostics consumer. Fix is straightforward


rcarriga:
Well that's a bit stupid :facepalm: Thanks for the fix!

PR #86, opened 2022-08-07
[RFC] Delay setting file type of summary buffers

HiPhish:
The standard Vim way of setting filetype-specific settings is to add them to an `ftplugin/my-filetype.{vim,lua}` file (`:h ftplugin`), but this does not in Neotest work with window-specific options like `'spell'`. The solution was to delay setting the file type until after the window exists. While at it I also moved the default settings to a dedicated file to save some lines of code and I added the `nospell` setting because I think we should have that as a default.

Personally I would remove the `norelativenumber` setting from the defaults because we already have `nonumber` in the defaults. If users want to turn numbers back on we should not be forcing absolute numbers on them. What is your opinion?

### DIFF
--- a/ftplugin/neotest-summary.vim
+++ b/ftplugin/neotest-summary.vim
@@ -1,1 +1,17 @@
-setlocal nomodifiable norelativenumber nonumber winfixwidth nospell
+if get(b:, 'did_ftplugin', v:false)
+	finish
+endif
+let b:did_ftplugin = v:true
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+setlocal winfixwidth
+setlocal nonumber
+setlocal norelativenumber
+setlocal nospell
+
+let b:undo_ftplugin = 'setlocal wfh< nu< rnu< spell<'
+
+let &cpo = s:save_cpo
+unlet s:save_cpo
--- a/ftplugin/neotest-summary.vim
+++ b/ftplugin/neotest-summary.vim
@@ -1,0 +1,1 @@
+setlocal nomodifiable norelativenumber nonumber winfixwidth nospell
--- a/lua/neotest/consumers/summary/init.lua
+++ b/lua/neotest/consumers/summary/init.lua
@@ -16,6 +16,7 @@ local function create_buf()
   end
 
   summary_buf = async.api.nvim_create_buf(false, true)
+  async.api.nvim_buf_set_option(summary_buf, 'modifiable', false)
   async.api.nvim_buf_set_name(summary_buf, buf_name)
   return summary_buf
 end
--- a/lua/neotest/consumers/summary/init.lua
+++ b/lua/neotest/consumers/summary/init.lua
@@ -16,13 +16,6 @@ local function create_buf()
   end
 
   summary_buf = async.api.nvim_create_buf(false, true)
-  local options = {
-    modifiable = false,
-    filetype = "neotest-summary",
-  }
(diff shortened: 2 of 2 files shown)


rcarriga:
Hey, thanks for raising :smile: 

> The standard Vim way of setting filetype-specific settings is to add them to an ftplugin/my-filetype.{vim,lua}

Yep this works well for traditional vim configs, and definitely agree that we should cater for this style of config. Delaying the file type setting seems like a good approach :+1: 

> While at it I also moved the default settings to a dedicated file to save some lines of code

I disagree with this change, I'd rather have the code that sets up the summary buffer also contain the buffer/window settings. Using the vim style is a departure from the standard, modularised approach used throughout this code base and I would not expect readers to know to check `ftplugin`.

> I added the nospell setting because I think we should have that as a default.

Very reasonable :+1: 

> Personally I would remove the norelativenumber setting from the defaults because we already have nonumber in the defaults. If users want to turn numbers back on we should not be forcing absolute numbers on them. What is your opinion?

Having `nonumber` and `relativenumber` still results in the number column being shown which is not the intended behaviour.

HiPhish:
> Having `nonumber` and `relativenumber` still results in the number column being shown which is not the intended behaviour.
OK, I did not know that turning on `relativenumber` shows the number column even if `nonumber` is set, that's pretty dumb. I never use relative numbers.

> I disagree with this change, I'd rather have the code that sets up the summary buffer also contain the buffer/window settings. Using the vim style is a departure from the standard, modularised approach used throughout this code base

Can you please elaborate? I really don't understand your reasoning, using `ftplugin` *is* the Vim/Neovim standard way. You have to write less code, it is automatic, and it is immediately discoverable to contributors. I mean, if you wanted to count the number of lines in a file would you write a custom Python script that opens a file, loops through the contents line by line, increments a counter and prints the result, or would you just run `wc -l myfile.txt` from the shell? I would certainly go for the latter because even one second spent is wasted when `wc` comes with every Linux install. In the same vein `ftplugin` will always work on every Neovim install for every user. And there is more to `ftplugin` that a user would expect: `:h undo_ftplugin`, disabling settings, disabling default mappings. See `:h filetype-plugins`, do you really want to re-implement all this yourself? If Vim script is the problem, then you can use Lua instead, or even have both a Vim script and a Lua file.

> and I would not expect readers to know to check `ftplugin`.

`ftplugin` is a well documented standard feature that has existed since before Neovim, if users don't know about it then that's really on them. You have to know the editor basics before you start adding plugins. Besides, where the options are set does not matter to users anyway, it is only relevant to contributors.

Personally, I had to go digging through the Lua code base to find where the options are set precisely because they where not set in the standard location. There is not much code so it was not a big waste of time, but it was still more effort than it should have been.

rcarriga:
On further thought, I'm OK with the window settings being placed in a `ftplugin` file but still want the `modifiable` setting set in the lua code. The window settings are purely to improve the UX, the `modifiable` setting is part of the functionality and is altered elsewhere.

Please restore the `modifiable` setting in the lua code and also let's place the settings on separate lines just for clearer version control.

HiPhish:
I also went ahead and implemented `b:did_ftplugin`, `b:undo_ftplugin` and added `cpo` guards in the script for the sake of completeness.

> but still want the `modifiable` setting set in the lua code. The window settings are purely to improve the UX, the `modifiable` setting is part of the functionality and is altered elsewhere.

Yes, that makes sense.

PR #100, opened 2022-08-28
feat: project configurations

rcarriga:
Allows users to configure settings per-project including test discovery
and adapters.

### DIFF
--- a/doc/neotest.txt
+++ b/doc/neotest.txt
@@ -150,6 +150,13 @@ Default values:
   }
 <
 
+neotest.setup_project()                                *neotest.setup_project()*
+               `neotest.setup_project`({path}, {project_config})
+Configure a project individually.
+Parameters~
+{path} `(string)`
+{project_config} neotest.CoreConfig
+
 
 ==============================================================================
 neotest.config                                                  *neotest.config*
--- a/doc/neotest.txt
+++ b/doc/neotest.txt
@@ -151,10 +151,10 @@ Default values:
 <
 
 neotest.setup_project()                                *neotest.setup_project()*
-               `neotest.setup_project`({path}, {project_config})
+           `neotest.setup_project`({project_root}, {project_config})
 Configure a project individually.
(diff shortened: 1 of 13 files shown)


akinsho:
@rcarriga would it be possible to add the ability to set up a project as a function call separate to the setup call. Reason being I'd ideally like to keep any specific of various projects out of my dotfiles and instead rely on local vimrc files that way inside of a project specific rc file I could do

```lua
--- ./project/.vimrc.lua which I load
require('neotest').configure_project({
  discovery = whatever
})
```

without having to do

```lua
require('neotest').setup({
 projects = {
  ['client/I/dont/want/to/share'] = ...,
  ['old/client/i/forgot/to/remove'] = ...,
 }
})
```

rcarriga:
Yeah that makes sense :+1: I've added the `neotest.setup_project` function, which requires a path. So for that use case, assuming you open nvim in the root dir, it would be:
```lua
require("neotest").setup_project(vim.loop.cwd(), {
  discovery = { enabled = false }
})
```

stevearc:
This is exactly the API I was looking for! Unfortunately, it doesn't seem to be working for me. I have this in one of my init files:
```lua
      neotest.setup_project('/Users/stevearc/work/repo', {
        discovery = { enabled = false },
        adapters = {
          my_custom_work_adapter({}),
          jest_adapter({
            cwd = jest_adapter.root,
          }),
        },
      })
```
But test discovery seems to not work. I'm opening vim in the `/Users/stevearc/work/repo` directory, and the test file will be in some subdir of that. I haven't done a full debug pass, but it looks like none of the adapter methods are even being called (e.g. `is_test_file`). I can look into this more, but wanted to check: is this config expected to work or did I mess it up somehow?

rcarriga:
I hadn't properly tested the adapter usage and missed some things. Please try now

stevearc:
Now it works!

PR #120, opened 2022-10-01
feat(output): open output for last test run

marcelbeumer:
Running `:lua require('neotest').output.open({ last_run = true })` will open output of last test run. 

I created this to support the workflow:

- Run a test with `:lua require("neotest").run.run()`.
- Edit implementation in a split, next to the test file.
- Run `:lua require("neotest").run.run_last()` to run last test while still in non-test buffer.
- Run `:lua require('neotest').output.open({ last_run = true })` to check it's output still in the non-test buffer.

Should not be feature creep, so if there's a better way to support this workflow, let me know!

### DIFF
--- a/lua/neotest/consumers/output.lua
+++ b/lua/neotest/consumers/output.lua
@@ -174,7 +174,10 @@ function neotest.output.open(opts)
   async.run(function()
     local tree, adapter_id
     if opts.last_run then
-      tree, adapter_id = run.get_last_run()
+      local position_id, last_args = run.get_last_run()
+      if position_id and last_args then
+        tree, adapter_id = client:get_position(position_id, last_args)
+      end
       if not tree then
         lib.notify("Last test run no longer exists")
         return
--- a/lua/neotest/consumers/output.lua
+++ b/lua/neotest/consumers/output.lua
@@ -1,7 +1,6 @@
 local async = require("neotest.async")
 local lib = require("neotest.lib")
 local config = require("neotest.config")
-local run = require("neotest").run
 
 local win, short_opened
 
@@ -174,7 +173,7 @@ function neotest.output.open(opts)
   async.run(function()
     local tree, adapter_id
     if opts.last_run then
-      local position_id, last_args = run.get_last_run()
+      local position_id, last_args = require("neotest").run.get_last_run()
       if position_id and last_args then
         tree, adapter_id = client:get_position(position_id, last_args)
       end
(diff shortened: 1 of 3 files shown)


lua/neotest/consumers/output.lua
```diff
@@ -171,7 +173,15 @@ function neotest.output.open(opts)
   end
   async.run(function()
     local tree, adapter_id
-    if not opts.position_id then
+    if opts.last_run then
+      local last_run = run.get_last_run()
```
rcarriga, line 177: Need to check if `last_run` is `nil`

lua/neotest/consumers/run.lua
```diff
@@ -13,6 +13,11 @@ local neotest = {}
 --- A consumer providing a simple interface to run tests.
 neotest.run = {}
 
+---@private
```
rcarriga, line 16: Similar comment, we can't treat this as private if another consumer is using it.

I think we should do the unpacking (and thus nil check) here as it'll be a more consistent API. 

Please also add a doc message with a description and return values, the Github Actions will generate a commit to update the docs with it.

rcarriga (changes requested):
Thanks for the PR! I definitely think this is a good feature to have :smile: Just a few small comments

lua/neotest/consumers/run.lua
marcelbeumer: Done. Still seems a bit strange to have a public `.run()` that returns a `tree, adapter_id`?

lua/neotest/consumers/run.lua
```diff
@@ -197,6 +197,16 @@ function neotest.run.adapters()
   return client:get_adapters()
 end
 
+--- Get last test run tree and adapter id.
+---@return neotest.Tree | nil, integer | nil
+function neotest.run.get_last_run()
```
marcelbeumer, line 202: @rcarriga: still seems a bit strange to have a public `.run()` that returns a `tree, adapter_id`? But have to admit I didn't spend enough time to really understand the code's data structures and API's properly..

rcarriga: Sorry I should have been clearer :facepalm: The function just return `position_id, last_args`, not the tree as yes you're right the tree is not really meant for general usage, just for consumers.

lua/neotest/consumers/output.lua
```diff
@@ -171,7 +172,16 @@ function neotest.output.open(opts)
   end
   async.run(function()
     local tree, adapter_id
-    if not opts.position_id then
+    if opts.last_run then
+      local position_id, last_args = require("neotest").run.get_last_run()
```
marcelbeumer, line 176: when doing a `require('neotest').run` on top of the file of a `neotest.run` here, I will get an error run being `nil`

rcarriga:
LGTM, thanks again for taking this on :grin:

Issue #110, opened 2022-09-12
[feature]: statusline integration


rebelot:
Hi, is it possible to retrieve a table of informations to be displayed in custom statuslines?
something like:

```lua
local function my_neotest_statuline_component()
    local dict = require'neotest'.status_dict
    return "OK " .. dict.passed .. "FAILED " .. dict.failed .. " / " .. dict.total
end
```

maybe it is already possible given the large API of neotest but I couldn't find a way to access tests results

rcarriga:
This is not yet possible, it would require a new consumer. Happy to take contributions :smile:

rebelot:
I'm trying to write a consumer, which kinda works so far. However, the client:get_results() method returns the results for single tests and for the entire test file and I haven't found a "comfortable" way to separate the two kind of results

rcarriga:
You probably want to iterate through the tree and get the results for only tests
```lua
local results = client:get_results(adapter_id)
local tree = client:get_position(root_id, { adapter = adapter_id })
for _, pos in tree:iter() do
  if pos.type == "test" then
     ...
  end
end
```
Results can come in very often when streamed so I'd recommend having some sort of caching to prevent a tonne of work being done. You could just mark the cache as invalid on a results event and then recalculate when the status is requested

rebelot:
```lua
    consumers = {
        statusline = function(client)
            return {
                statusline = function()
                    local adapters = client:get_adapters()
                    local statuses = {}
                    local root_id = vim.api.nvim_buf_get_name(0)
                    for _, adapter_id in ipairs(adapters) do
                        local results = client:get_results(adapter_id)
                        local tree = client:get_position(root_id, { adapter = adapter_id })
                        if not tree then
                            return
                        end
                        for _, pos in tree:iter() do
                            if pos.type == "test" then
                                local id = pos.id
                                if results[id] then
                                    table.insert(statuses, results[id].status)
                                else
                                    table.insert(statuses, "none")
                                end
                            end
                        end
                    end
                    local status_dict = { total = 0, failed = 0, passed = 0 }
                    for _, s in ipairs(statuses) do
                        if s == "failed" then
                            status_dict.failed = status_dict.failed + 1
                        elseif s == "passed" then
                            status_dict.passed = status_dict.passed + 1
                        elseif s == "none" then
                            status_dict.total = status_dict.total + 1
                        end
                    end
                    status_dict.total = status_dict.total + status_dict.failed + status_dict.passed
                    return status_dict
[...]

PR #134, opened 2022-10-28
feat(consumers): add state consumer (statusbar integration)

Davincible:
Added a consumer that serves as an internal state cache, which can be consumed for use in e.g. status bars

My status bar example:
image

Ref: #110

### DIFF
--- a/lua/neotest/config/init.lua
+++ b/lua/neotest/config/init.lua
@@ -2,7 +2,7 @@
 ---@toc_entry Configuration Options
 
 local function define_highlights()
-  vim.cmd([[
+	vim.cmd([[
   hi default NeotestPassed ctermfg=Green guifg=#96F291
   hi default NeotestFailed ctermfg=Red guifg=#F70067
   hi default NeotestRunning ctermfg=Yellow guifg=#FFEC63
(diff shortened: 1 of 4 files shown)


Davincible:
Probably still needs some docs

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,198 @@
...
+    end
+  end
+  return nil
+end
+
+---Get status count (passed | failed | skipped | unknown)
+---@param path_query string
+---@param opts table
+---       :fuzzy use string.match instead of direct comparison for key
+---       :status get count for status
```
rcarriga, line 161: If this is required, it should be it's own parameter and we should assume it's provided

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,198 @@
+local au_result = "NeoTestStatus"
```
rcarriga, line 1: Maybe `NeotestResult` would be a more semantic name?

lua/neotest/config/init.lua
```diff
@@ -109,122 +109,125 @@ define_highlights()
...
-    signs = true,
-  },
-  run = {
-    enabled = true,
-  },
-  jump = {
-    enabled = true,
-  },
-  projects = {},
+	log_level = vim.log.levels.WARN,
```
rcarriga, line 112: Please run the `scripts/style` script :smile:

rcarriga (changes requested):
Thanks for the PR! Very nice :smile: Mainly issues around public facing APIs just to nail them down because we don't want breaking changes in future

Davincible:
Thanks for the review :) I'll have a look

lua/neotest/consumers/state/init.lua
```diff
@@ -0,0 +1,68 @@
...
+---Alias for client:get_adapters to get registerd adapters
+---@return string[]
+function neotest.state.get_adapters()
+  return internal_state:get_adapters()
+end
+
+---Get back results from cache
+---@param path_query string
+---@param opts table
+---       :fuzzy use string.match instead of direct comparison for key
```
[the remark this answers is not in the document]
Davincible: LSP doesn't like this but I've updated them

image

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,198 @@
...
+    end
+    self._cache[path] = count
+  end
+end
+
+---Receive result from event and add to internal state
+---@param adapter_id string
+---@param results table<string, neotest.Result>
+function state:_update_results(adapter_id, results)
+  self._running[string.match(adapter_id, "^[^:]*:(.*)")] = nil
```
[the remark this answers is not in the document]
Davincible: I'm extracting the path name here from the provided adapter_id format. `client:get_adapter(adapter_id)` doesn't yield anything

lua/neotest/consumers/state/init.lua
```diff
@@ -0,0 +1,68 @@
...
+end
+
+---Check if there are tests running
+---Optionally, provide a "<adapter_id>:<file_path>" argument to filter on.
+---If no argument is provided, all running processes will be returned.
+---If nil is returned your adapter_id found no match, otherwise an emtpy
+---table is returned.
+---@param adapter_id string optional
+---@param opts table
+---       :fuzzy use string.match instead of direct comparison for key
```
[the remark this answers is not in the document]
Davincible: to match between things like `/home/abc/project/sub/file.go` and `/home/abc/project/`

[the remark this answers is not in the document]
Davincible: ^^ This is especially needed to fetch the running state, as the latter format is reported for suites.

Davincible:
@rcarriga anything I still need to change?

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,208 @@
...
+---Process results and count pass/fail rates, save to cache
+---Passed, Failed, Skipped
+---@param tree neotest.Tree|nil
+function state:_process_result(tree, results)
+  if not tree then
+    return
+  end
+  for _, pos in tree:iter() do
+    if pos.type == "test" then
+      if results[pos.id].status then
```
wookayin, line 60: I get `internal.lua:60: attempt to index a nil value` an error in this line.

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,198 @@
...
+end
+
+---Update status for specific test
+---@param path string
+---@param status string
+function state:_set_status(path, name, status)
+  if self._status[path] == nil then
+    self._status[path] = {}
+  end
+  self._status[path][name] = status
```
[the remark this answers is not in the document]
rcarriga: No names are not guaranteed to be unique. For example in the rspec adapter, it uses line numbers in the IDs because the names can be anything including empty. Also for adapters where names do have to be unique, it's only unique in their namespaces, e.g. in python two test classes in the same file can define a test with the same name

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,208 @@
...
+end
+
+---Update the cache status count
+function state:_update_cache()
+  local total = { passed = 0, failed = 0, skipped = 0, unknown = 0 }
```
rcarriga, line 80: `unknown` is not a valid status, tests are treated as unknown if they have no result recorded and are not running so this will always be 0

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,208 @@
...
+end
+
+---Check if there are tests running
+---If no options are provided, all running processes will be returned.
+---If nil is returned your adapter_id found no match, otherwise an empty
+---table is returned.
+---@param opts table
+---@field adapter_id string Optionally, provide a "<adapter_id>:<file_path>" argument to filter on.
+---@field fuzzy string use string.match instead of direct comparison for key
+---@return table<string, string> | string[] | nil
```
wookayin, line 128: It is not clear what is the expected return type of this function. In which case can it return nil or a list(table) string? It is not mentioned in the docstring.

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,208 @@
...
+function state:running(opts)
+  opts = opts or {}
+  if not opts.adapter_id or (type(opts.adapter_id) ~= string and #opts.adapter_id == 0) then
```
rcarriga, line 131: Type annotation just says it can be a string, not a list

lua/neotest/consumers/state/internal.lua
```diff
@@ -0,0 +1,208 @@
...
+    then
+      return value
+    end
+  end
+  return nil
+end
+
+---Get back results from cache. Fetches all results by default.
+---@param opts table
+---@field status string fetch count for specific status (passed | failed | skipped | unknown)
```
wookayin, line 147: what about "runnning" status?

lua/neotest/consumers/state/internal.lua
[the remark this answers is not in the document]
rcarriga: > I'm extracting the path name here from the provided adapter_id format.

Yes but we shouldn't care about the adapter ID structure, it's not a documented feature and can change in future. 

> client:get_adapter(adapter_id) doesn't yield anything

Can you please check that again, it's a commonly used function throughout the code so if it doesn't work then there's a bug somewhere :sweat_smile:

wookayin (changes requested):
Interesting feature. I also have been in need of similar feature; thanks for your work. Here's my twenty cents:

- For me, however, it does not work. When the test runs (or has completed), an error throws and due to the error, all the results are wrong. For example, a completed task is reported as still running and `state.get_status()` returns nothing (i.e., an empty table).

- I think the consumer name `state` is awkward. I feel it would be more natural to call it like a `status` consumer, but we have a conflict with an existing passive consumer. `state` sounds weird because it feels like the internal state of a client, but what we really want to extract here is the **status** of a test case.

lua/neotest/consumers/state/init.lua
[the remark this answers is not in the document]
rcarriga: But I don't understand why that would be needed? If `/home/abc/project/` is running then `/home/abc/project/sub/file.go` should report as running directly from the client already.

rcarriga:
Sorry for not being quicker to respond, it's quite a large PR so I need put time aside to review.

My main concern right now, is that there's a bit too much magic or "help" going on such as the fuzzy matching or the
different return types based on different scenarios (nil if no matching adapter ID found or empty list if matched but no
positions found for example). These features are very opinionated and I'd like to avoid them if possible for now.

The core goal of providing a cached state to read from is brilliant and I'm very eager to get it in! I don't have time
to complete a full review but wanted to leave you with my current thoughts.

rcarriga:
> I think the consumer name state is awkward. I feel it would be more natural to call it like a status consumer, but we have a conflict with an existing passive consumer. state sounds weird because it feels like the internal state of a client, but what we really want to extract here is the status of a test case.

I'd disagree with this because it leaves it open for extra states in future. For example it's returning the known adapter IDs, which the run consumer does but it makes much more sense in a general "state" consumer IMO

Davincible:
Thanks for the reviews! I'll have a look at them.

> Sorry for not being quicker to respond, it's quite a large PR so I need put time aside to review.

Of course, I understand, no problem :) 

> My main concern right now, is that there's a bit too much magic or "help" going on

I had deliberately created the API this way, as I could either create different functions for each small variation (e.g. `get_status`, `get_status_all`), which in my opinion unnecessarily increases the API surface as it could also be a simple option.

lua/neotest/consumers/state/internal.lua
Davincible: Currently yes. I had put this in as future work. I thought it could be nice to also discover unrun tests in the future, so you can display them as well. As I personally didn't care too much for it yet I haven't spend the time to implement it

Davincible:
@wookayin I'm not sure how you implemented it, but since there are no docs yet as reference here is my implementation:

Fetch count per status (pass/fail):
https://github.com/Davincible/dotfiles/blob/4b03eec730bb0ff19451a11ccf30311ad789765b/nvim/lua/uno/heirline/astro/status.lua#L509-L523

Display the current running status with animation:
https://github.com/Davincible/dotfiles/blob/4b03eec730bb0ff19451a11ccf30311ad789765b/nvim/lua/uno/heirline/astro/status.lua#L530-L546

Check if any test results are present at all, used as condition:
https://github.com/Davincible/dotfiles/blob/4b03eec730bb0ff19451a11ccf30311ad789765b/nvim/lua/uno/heirline/astro/status.lua#L696-L700

Check if neotest is currently running, used as condition:
https://github.com/Davincible/dotfiles/blob/4b03eec730bb0ff19451a11ccf30311ad789765b/nvim/lua/uno/heirline/astro/status.lua#L705-L709

Style definition of status:
https://github.com/Davincible/dotfiles/blob/4b03eec730bb0ff19451a11ccf30311ad789765b/nvim/lua/uno/heirline/astro/status.lua#L993-L1029

Style definition of running:
https://github.com/Davincible/dotfiles/blob/4b03eec730bb0ff19451a11ccf30311ad789765b/nvim/lua/uno/heirline/astro/status.lua#L1035-L1049

lua/neotest/consumers/state/internal.lua
Davincible: Comment here is slightly outdated, but:

* If tests are currently running, and you supply no arguments, you gets something like this:
  ```lua
  {                                                                                                                                                                                                  
    ["/home/tyler/Launchpad/Micro/v5-stuff/http-poc"] = {
      adapter = "neotest-go",
      path = "/home/tyler/Launchpad/Micro/v5-stuff/http-poc"
    }  
  }
  ```
* If no tests are running, or no match found, it returns `nil`

You can add a query to get only the nested table element back. E.g.:
```lua
local file = "/home/tyler/Launchpad/Micro/v5-stuff/http-poc/some/pkg/file.go"
local running = require("neotest").state.result({adapter_id = file, fuzzy = true})
```

@rcarriga this illustrates the fuzzy field, because I want to check if any tests are running for my current file, I somehow need to match the package.

lua/neotest/consumers/state/internal.lua
Davincible: You use the `running` method

lua/neotest/consumers/state/internal.lua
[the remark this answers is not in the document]
Davincible: If I put this in `_update_results`

```lua
  vim.notify(vim.inspect({
    adapter_id = adapter_id,
    adapter = { self._client:get_adapter(adapter_id) },
  }))
```

```lua
{
  adapter = {},
  adapter_id = "neotest-go:/home/tyler/Launchpad/Micro/v5-stuff/http-poc"
}
```

rcarriga:
> I had deliberately created the API this way, as I could either create different functions for each small variation (e.g. get_status, get_status_all), which in my opinion unnecessarily increases the API surface as it could also be a simple option. But I'm open to suggestions here ofcourse

I agree that the API shouldn't have lots of little functions but I think instead of trying to accommodate for those use cases in one function, it's better to expose one simple function that users can build on top of to do what they want.

I'm unclear on what exactly your goals are and so I think outlining the use cases and how they would be implemented would be a good start.

For me, the main use cases I see are 
1. Getting the counts of all the statuses
2. Finding out what is running
3. Knowing how many positions are running
It'd be useful to have those at a global level and at a per-buffer level

With those use cases in mind, a simple API to do that might look like the following:
```lua
---@return string[]
function neotest.state.get_adapters()
  ...
end

---@class neotest.state.StatusState
---@field total integer
---@field passed integer
---@field failed integer
---@field skipped integer

---@param adapter_id string
---@param opts table
---@field buffer? integer Returns statuses for this buffer
---@return neotest.state.StatusState
function neotest.state.get_status_counts(adapter_id, opts)
  ...
end

---@class neotest.state.RunningState
---@field root string ID of position that was run
---@field total integer Number of running positions under the root

---@param adapter_id string
---@param opts table
---@field buffer? integer Returns running position IDs for this buffer
---@return neotest.state.RunningState[]
function neotest.state.get_running(adapter_id)
  ...
end
```

I believe this covers all of the use cases that you have in your examples, but please let me know of any other requirements.

Requiring the adapter ID means that we don't have to do any odd changing of return types depending on the arguments,
while it doesn't add much complexity to the usage, it's just a simple loop for users.


In terms of implementation, on the neotest events all corresponding state will be cleared and the global levels will be refreshed.
Buffer level states will be only be calculated when first requested and then cached for later use.
It doesn't need to be super efficient, we don't need to be smart about the state clearing (e.g. trying to only wipe the
state of buffers that have changed) as the once off calculation won't be very heavy.

What do you think?

rcarriga:
@Davincible are you still interested in continuing this PR? If not I may have a stab at the API proposed above

Davincible:
@rcarriga sorry for the radio silence. I am.

I agree with the goals. However just using adapter ID is an oversimplification, as they don't always match up with what you are trying to query. See [this comment](https://github.com/nvim-neotest/neotest/pull/134#discussion_r1014670472) for more detail.

I agree that perhaps having so many options makes it too complex. We could make the fuzzy search the default behavior.

rcarriga:
Not a problem! :smile:

Your comment says that fuzzy searching is necessary because the `running` function returns only the root ID and that
might not necessarily be the file path. That's just an implementation issue. What I'd like to know is if fuzzy searching
is solving an actual use case and is it the best solution? Right now it just seems to be a way to compensate for a
lacking API.

This is why I was asking what are the use cases that you're trying to fulfill, because I just don't know exactly what
they are. The API I've proposed is designed to make querying for info on a buffer level as easy as possible which I
believe is what you're also trying to do. Is there something that it doesn't handle?

rcarriga:
Cool :smile: What about the proposed API? Are you happy with it as is or do you have other ideas?

> thus the "run" function will merely receive information of the project path, and not the adapter id,

The first argument to the run event is the adapter ID. However I think we should maintain a cache of the trees emitted during discovery, because then we can lookup those trees in the exposed API functions without doing potentially async calls in the client.

> and when running a suite (provided as a specific path), we could then change the state of all sub-paths to running.

This could be problematic when multiple parents of a position are running depending on implementation. Not sure it will be, just want to point it out.

weilbith:
May I ask about the current status of this PR? Can I help somehow?

I finally made the transition from Ultest to Amritsar, but having this missing bugs me.

rcarriga:
As this PR hasn't been updated in a while, I gave it a quick stab to get something to fulfil the needs discussed.
I've got it in a branch (#217) to test for now, so please let me know if anything isn't working or it doesn't cover the requirements :smile:

I actually just put the running count in the status table so that the count of `statuses["total"] == sum(count for status, count in statuses if status != "total"` so I didn't include the `get_running` function.

Davincible:
Hey, apologies for not working on it further. I wanted to continue but haven't found the time. 

I haven't had a look at your PR @rcarriga yet, but as long as it properly takes into account the different cases of running full test suites on an entire project vs single files or single tests within a file it should be good. Those were the biggest issues I was still facing throughout the usage of my PR here.

weilbith:
I just quickly read over the code. I'm not sure if I missed it and it is possible. But it would be great to also get the positions of the tests. So I can say (the filter is supported) I want to know all positions (row and column [span?]) of this buffer and the related result.
I'm not sure if this makes sense to be part of this consumer, but also don't really see the meaning to have this tiny thing as another consumer. I have some "cool" ideas for some plugins to build on top of this information.

rcarriga:
No problem @Davincible! Thanks for all the work you put in :smile: 

@weilbith It definitely makes sense, I can add something for that.

Issue #143, opened 2022-11-09
[Feature] summary: support marking/running tests in visual/visual line mode


OddBloke:
This would make marking/running multiple contiguous tests much quicker.

PR #144, opened 2022-11-09
feat(summary): map mark/run/stop keybindings in visual mode

OddBloke:
These, predictably, perform the given action across the tests in the selected visual range.

This commits adds a summary.map_in_visual config option: actions listed there will have their configured keys mapped in visual mode.

Fixes: #143

### DIFF
--- a/lua/neotest/consumers/summary/canvas.lua
+++ b/lua/neotest/consumers/summary/canvas.lua
@@ -117,6 +117,7 @@ function Canvas:render_buffer(buffer)
     return false, "Window not found"
   end
 
+  local map_in_visual = { "expand", "expand_all", "mark", "run", "stop" }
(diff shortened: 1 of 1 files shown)


rcarriga:
Very nice! Though don't think I see the point of exposing which actions are mapped, instead we just allow it for all of the ones where it makes sense. I think adding `expand` and `expand_all` would cover it.

OddBloke:
> Very nice! Though don't think I see the point of exposing which actions are mapped, instead we just allow it for all of the ones where it makes sense. I think adding `expand` and `expand_all` would cover it.

Done! I started off making the bindings separately configurable, then realised that would be annoying, but didn't follow that to its logical conclusion.

rcarriga:
Fantastic, I really like this change! Thanks again for the PR :tada:

PR #145, opened 2022-11-09
feat(summary): add any_marked function to API

OddBloke:
This enables API consumers (such as user config) to modify their behaviour based on whether or not there are marked tests.

### DIFF
--- a/lua/neotest/consumers/summary/init.lua
+++ b/lua/neotest/consumers/summary/init.lua
@@ -256,6 +256,20 @@ function neotest.summary.run_marked(args)
   end
 end
 
+--- Are there any marked tests?
+---
+--- @return boolean
+function neotest.summary.any_marked()
+  for _, component in pairs(components) do
+    for _, marked in pairs(component.marked) do
+      if marked then
+        return true
+      end
+    end
+  end
+  return false
+end
+
 --- Clear all marked positions
 ---@param args table
 ---


OddBloke:
With this, I can auto-run marked tests, if any, or else the whole file:

```lua
local test_whole_file = function()
    neotest.run.run(vim.fn.expand("%"))
end

local test_marked_else_whole_file = function()
    if neotest.summary.any_marked() then
        neotest.summary.run_marked()
    else
        test_whole_file()
    end
end

local autotestGrp = vim.api.nvim_create_augroup("initlua-neotest-autorun", { clear = true })

vim.api.nvim_create_autocmd("BufWritePost", {
  callback = test_marked_else_whole_file,
  group = autotestGrp,
})
```

rcarriga:
I like the idea, but instead how about we provide a `marked` function which returns a `table<adapter id, position_id[]>`? Then your code can just check if the result is empty and others can do whatever mad things they want with them? :smile:

OddBloke:
I was reflecting on this further: I see `run_tree` returns a table of results. Perhaps those results could be returned all the way out to `run_marked`, then I could do this (pseudocode):

```
results = neotest.summary.run_marked()
if <length of results == 0> then
    test_whole_file()
end
```

This might be more generally useful than this API function, too?

> I like the idea, but instead how about we provide a `marked` function which returns a `table<adapter id, position_id[]>`? Then your code can just check if the result is empty and others can do whatever mad things they want with them? smile

I did consider that: I _think_ this implementation is slightly faster than that would be, as it only has to iterate as far as the first result? (Perhaps I'm missing knowledge about Lua iterators?) It's not a huge difference, really, so I'd be happy to switch if we do pursue the API function.

rcarriga:
> I was reflecting on this further: I see run_tree returns a table of results. Perhaps those results could be returned all the way out to run_marked,

`neotest.run.run` is a synchronous function that kicks off an async function so you'd have to use a callback which I don't think would be very nice. Also it might work for your use case here but exposing the marked positions makes the summary window much more extensible for other use cases as well as yours. 

> I did consider that: I think this implementation is slightly faster than that would be, as it only has to iterate as far as the first result? (Perhaps I'm missing knowledge about Lua iterators?) It's not a huge difference, really, so I'd be happy to switch if we do pursue the API function.

Yeah I wouldn't worry about speed here at all, the number of marked positions is almost always tiny. Even if you were marking thousands of positions at once, luajit is still very good at creating lists so the time difference would be tiny. The more flexible API definitely beats any concern here :+1:

OddBloke:
Done this locally, resubmitting in a new PR.

OddBloke:
> Done this locally, resubmitting in a new PR.

#156

PR #146, opened 2022-11-10
feat(summary): add summary_toggle event

OddBloke:
This is emitted when the summary window is opened (`true`) or closed (`false`), allowing other consumers to perform actions when the summary window is toggled.

In my case, I want to enable automatic running of tests on write only when the summary window is open: I will likely tie other behaviour in my configuration to this state too, as a proxy for "concentrating on tests".

(I investigated adding events to consumers (so that the event could be defined within `summary/init.lua` itself), but it was getting complicated: it would require extending the Client interface, and most consumers won't need it. If you'd prefer a more general solution, I'm happy to hack at it some more!)

### DIFF
--- a/lua/neotest/client/events/init.lua
+++ b/lua/neotest/client/events/init.lua
@@ -6,11 +6,15 @@ local M = {}
 ---@alias neotest.Event "discover_positions" | "run" | "results" | "test_file_focused" | "test_focused"
 ---@class neotest.Events
 local NeotestEvents = {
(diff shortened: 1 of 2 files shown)


rcarriga:
I like the idea of using events for consumers but would want to avoid them using the client.

Instead we could use autocmds which are already being used in #134. They are also more user friendly so people don't need to write consumers if they want to listen to them

PR #149, opened 2022-11-11
feat(treesitter): refactor generic querying to separate functions

OddBloke:
This allows it to be used by other parts of the code, or plugins, who need to perform a query:

```
local query = "<a treesitter query>"
local content = lib.files.read(path)
local ts_root, lang = lib.treesitter.get_parse_root(path, content, { fast = true })
local built_query = lib.treesitter.normalise_query(lang, query)
local has_match = built_query:iter_matches(ts_root, content)() ~= nil
```


rcarriga:
Looks great!

PR #154, opened 2022-11-16
Add adapter for Pest (PHP)

theutz:
Simple change to the README to list an adapter for [PestPHP](https://pestphp.com).


rcarriga:
Nice thanks for the PR! :smile: 

Had a quick look and FYI instead of [matching file path prefixes for a certain directory](https://github.com/theutz/neotest-pest/blob/main/lua/neotest-pest/init.lua#L22), you can define a `filter_dir` function so that only those directories are searched in the first place https://github.com/nvim-neotest/neotest/blob/master/lua/neotest/adapters/interface.lua#L17

theutz:
Thanks for the merge and the excellent tip! I implemented it as you indicated, and it seems to speed things up, too!

PR #155, opened 2022-11-17
Add return annotation for `filter_dir` interface

theutz:
Thought this might help future devs more easily discover the joy of the `filter_dir` interface you showed me in #154. :)


rcarriga:
Ah thank you!

PR #156, opened 2022-11-17
feat(summary): add marked function to API

OddBloke:
This enables API consumers (such as user config) to query for marked tests, enabling them to change their behaviour when there are.


rcarriga:
Brilliant thanks!

Issue #157, opened 2022-11-17
[BUG] Sidebar briefly shows "No tests found" (instead of "Parsing tests")


OddBloke:
I've seen this reliably when sleeping 5s for my pytest debugging, but I've just reproduced it with trunk of neotest and neotest-python. I believe the issue is that https://github.com/nvim-neotest/neotest/blob/36aa842f4f344929929110b8f22cbed2aeb4b83c/lua/neotest/consumers/summary/summary.lua#L93-L95 presumes that `has_started()` means "has completed parsing" whereas https://github.com/nvim-neotest/neotest/blob/36aa842f4f344929929110b8f22cbed2aeb4b83c/lua/neotest/client/init.lua#L342-L344 sets `self._started` (which `has_started()` returns) well before the first parse is performed: https://github.com/nvim-neotest/neotest/blob/36aa842f4f344929929110b8f22cbed2aeb4b83c/lua/neotest/client/init.lua#L436

Given that `self._started` is used in `ensure_started`, moving it after all parsing is complete (which could take seconds in a large project, if shelling out is involved) seems like the wrong move. Perhaps a new interface to check if the initial parse has occurred (implemented as `self._initial_parse_complete = true` after discovery + a getter)?

rcarriga:
I actually don't like the `has_started` interface at all, it was just a bit of a hack. I've taken the chance to introduce the `starting` and `started` events which are more in line with the rest of the client usage. If you could try out the branch in #159 and let me know if you see any issues that'd be great.

This would actually be the first breaking change neotest has made :sweat_smile:  but it's a consumer facing change so don't think it should affect many people.

OddBloke:
Just tried out #159, looks good to me!

PR #159, opened 2022-11-18
feat: started and starting events

rcarriga:
BREAKING CHANGE: Removes the `has_started` and `ensure_started` client methods

See #157


OddBloke (approved):
Fixes #157 for me locally, and the code LGTM (insofar as my burgeoning Lua-fu can tell!)

PR #160, opened 2022-11-18
fix: run fails if there are unloaded buffers

corymhall:
`vim.api.nvim_buf_get_name()` will throw an error if run on an unloaded buffer. I'm using `noice.nvim` which I think creates some unloaded buffers.

This PR adds a check to ensure the buffer is loaded before trying to get the name.


rcarriga:
Thanks for the fix!

Issue #136, opened 2022-10-30
Allow docking the output via defaults keybinds (e.g `<c-w>J`) without automatically closing it when leaving the window


DasOhmoff:
Hello 👋, thanks again for this nice plugin :)

Sometimes it is very convenient to see the output test result right next to you while you are editing some code, without having to reopen it all the time.

One can use the `<c-w>J`, `<c-w>K`, `<c-w>H` and `<c-w>L` mappings to be able to dock floating windows as normal windows. That is why I do so with the test outputs. Initially, the window gets docked, but as soon as I leave the output window, the output window automatically closes.

To make this work, I think the only thing needed here is to not automatically close the output floating window when it is left. The user can close the window himself if he wants to, by for example using `q`

It would be very nice and convenient to make this happen.

Thank you for your help :)

rcarriga:
I've added an `auto_close` option for the open function :smile:

DasOhmoff:
Thats perfect! Thank you! :D

wookayin:
This is a great feature @rcarriga, but it appears that somehow you forgot to add a `auto_close` field into the default config (`neotest/config/init.lua`).

rcarriga:
FYI @DasOhmoff the `auto_close` setting should no longer be needed for your use case thanks to @wookayin :smile: Once a window is moved to a non-floating window, it will stay open

DasOhmoff:
Ah ok nice. Thanks :)

PR #168, opened 2022-11-22
fix(float): Do not auto-close window if it is no longer a floating win

wookayin:
This would be a better solution for #136; when a neotest output window
is moved (initialized as a floating window) as normal windows,
the window should be no longer automatically closed.


rcarriga:
Thanks for the PR!

Issue #147, opened 2022-11-11
[Feature] Support child tests which cannot be focused


OddBloke:
In https://github.com/nvim-neotest/neotest-python/pull/36, I'm working on introducing support for pytest parameterization, which produces multiple test instances for each test definition. These test instances do not have a unique location in the codebase: they can be generated by the cross-product of multiple different parameter declarations (which do not necessarily even have to be in the same test file). It follows that they should never be selected/focused by location: only their parent test should be.

I need some support from neotest core for this use case: without any changes, the last test instance discovered for a given test will be focused in the summary view, and only that test will be run by `neotest.run.run`. The correct UX in these cases is for the parent test to be focused, and the parent test (and so all instances) to be run.

OddBloke:
My first thought for a proper implementation here was to indicate that a test can't be focused by setting `range` to `nil` (as opposed to setting it to the same as the parent). Code within neotest core would switch from reading `range` directly to finding the closest node in the tree with non-`nil` `range` and using its value. I've pushed up a rough partial PoC of this idea: https://github.com/nvim-neotest/neotest/commit/3df6db746a48313fef2541289b113a75a5ba23cc

However, I realised that it's only cases where we're mapping from a location to a test that we need to change the current behaviour: if we're annotating lines with diagnostics or statuses, or jumping to them, then we can use either the child or parent's range, and the outcome will be the same. Instead of having to traverse the tree every time we encounter a child test (albeit just one node up, generally), we can instead have tests indicate whether or not they can be selected when mapping a location:

```diff
--- a/lua/neotest/client/init.lua
+++ b/lua/neotest/client/init.lua
@@ -122,10 +122,12 @@ function NeotestClient:get_nearest(file_path, row, args)
   local nearest
   for _, pos in positions:iter_nodes() do
     local data = pos:data()
-    if data.range and data.range[1] <= row then
-      nearest = pos
-    else
-      return nearest, adapter_id
+    if data.can_focus then
+      if data.range and data.range[1] <= row then
+        nearest = pos
+      else
+        return nearest, adapter_id
+      end
     end
   end
   return nearest, adapter_id
diff --git a/lua/neotest/lib/treesitter/init.lua b/lua/neotest/lib/treesitter/init.lua
index 11ddc80..c80ee72 100644
--- a/lua/neotest/lib/treesitter/init.lua
+++ b/lua/neotest/lib/treesitter/init.lua
[...]

PR #172, opened 2022-11-23
feat(neotest): support range-less "child" tests

OddBloke:
Not all tests have a unique range which applies to them.  For example, in pytest, a test function can be parametrized to produce multiple different test instances for the same range.  neotest currently assumes that any cursor position in a file maps to a single test, and selects the deepest-nested such test: this results in the last test instance for a test function being focused, instead of the whole test: the only way to run other (or all) instances is via the summary view.

This commit changes neotest's behaviour, by introducing support for "child" test positions: these are simply test positions with `range` set to `nil`.  This commit updates all[0] direct accesses of `range` on positions with calls to `Tree:closest_value_for("range")`: this will traverse up the parents of a node, returning the first non-`nil` `range` value.  Child tests can be run via the summary view (directly, or via marks), but any operations based on cursor position in a buffer will operate on the parent test.

Fixes: #147

[0] With the exception of `positions.contains()`, which doesn't handle
    _test_ positions.

### DIFF
--- a/lua/neotest/types/tree.lua
+++ b/lua/neotest/types/tree.lua
@@ -156,6 +156,29 @@ function neotest.Tree:iter_parents()
   end
 end
 
+--- Fetch the first node ascending the tree (including the current one) with the
+--- given data attribute e.g. `range`
+---@param data_attr string
+---@return neotest.Tree | nil
+function neotest.Tree:closest_node_with(data_attr)
+  if self:data()[data_attr] ~= nil then
+    return self
+  end
+  for parent in self:iter_parents() do
+    if parent:data()[data_attr] ~= nil then
+      return parent
+    end
+  end
+end
+
+--- Fetch the first non-nil value for the given data attribute ascending the
+--- tree (including the current node) with the given data attribute.
+---@param data_attr string
+---@return any | nil
+function neotest.Tree:closest_value_for(data_attr)
+  return self:closest_node_with(data_attr):data()[data_attr]
+end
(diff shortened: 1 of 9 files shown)


lua/neotest/types/tree.lua
```diff
@@ -144,6 +144,21 @@ function Tree:iter_parents()
   end
 end
 
+function Tree:closest_node_with(data_attr)
```
rcarriga, line 147: Let's add a small docstring describing what this is doing, because closest could be taken to include siblings.

rcarriga (changes requested):
Looks great! Just a small doc and then good to go

rcarriga:
Actually ran into an issue when testing with your pytest PR. 
https://github.com/rcarriga/neotest/blob/283b71dcdc5aebca24a848da0fb23581596de8c6/lua/neotest/client/runner.lua#L236
If a range-less position is run from the summary, it errors when checking if the root contains the node.
A quick fix would be to do something like this:
```lua
  local all_position_ids = {}
  for _, pos in tree:iter() do
    all_position_ids[pos.id] = true
  end

  local function propagate_result_upwards(node)
    for parent in node:iter_parents() do
      local parent_pos = parent:data()
      if not all_position_ids[parent_pos.id] then
        return
      end
```
but open to other routes too

rcarriga:
Made the changes myself just to get this in as another PR is waiting on it, thanks again :smile:

PR #185, opened 2022-12-15
feat: publish releases to luarocks

mrcjkb:
neotest is a common dependency across Neovim plugins. Using luarocks may alleviate the need for users to specify their plugins' dependencies in their plugin manager. (e.g., vim-plug or packer).

See also:
https://teto.github.io/posts/2021-09-17-neovim-plugin-luarocks.html


MrcJkb:
If this MR is accepted, a package on [www.luarocks.org](http://www.luarocks.org/) will need to be added, which can be done from only one account. See https://github.com/luarocks/luarocks-site/issues/173.

rcarriga:
Very cool thanks for the PR! One question on the copied directories, do you know if this is this handled by packer by adding them to the RTP? I'd be worried that if a user installs it from luarocks, they wouldn't see the docs

MrcJkb:
> Very cool thanks for the PR! One question on the copied directories, do you know if this is this handled by packer by adding them to the RTP? I'd be worried that if a user installs it from luarocks, they wouldn't see the docs

Good question.
Packer uses hererocks under the hood, and [there's an open issue to change that](https://github.com/wbthomason/packer.nvim/issues/526).
[This blog post](https://teto.github.io/posts/2022-06-22-neovim-plugin-luarocks-2.html) seems to suggest that packer does (or will?) copy directories in the `copy_directories` directive to the installation root.

rcarriga:
OK just wanted to check thank you. One thing, I've tried adding nvim-treesitter as a dependency but it fails with `Error fetching file: Failed downloading http://github.com/nvim-treesitter/nvim-treesitter/archive/vscm.zip - failed downloading http://github.com/nvim-treesitter/nvim-treesitter/archive/vscm.zip`, is that an issue specific to nvim-treesitter? Is that supposed to resolve to something? The corresponding URL in the rockspec in this PR doesn't resolve to anything either

rcarriga:
It also seems the version of plenary that is available is outdated (missing the `filetype` module) and so isn't compatible. It seems like the `scm` version is a manually uploaded version, not an automatic zip of the latest code as I had first assumed. I think for a proper implementation, I'll need to alter the existing release workflow to push a luarock release. We'll also need the other libraries to do the same before this can be used by users

MrcJkb:
Yup. I'm facing similar issues with one of my repos that I'm trying to make compatible with luarocks. Let's mark this as draft for now, and I'll get back to it when I know more and have a solution.

teto:
>  It seems like the the scm version is a manually uploaded version,

The scm version installs the one from the "source control manager", git usually. So it should install the latest version. Problem is you have to tell luarocks to install dev/"scm" versions otherwise it doesn't see them. 
Looks like there are issues with the nvim-treesitter rockspec (and the git tags appeared after the rockspec, starting with neovim 0.8). I've tried to fix the issues to upload a stable rockspec to luarocks:
https://github.com/teto/nvim-treesitter/pull/new/release-rockspec-0.8.1

but I still have an issue: with the "builtin" build type of luarocks, luarocks tries to  compile the test files of nvim-treesitter and some of them are broken on purpose which makes the install so I am not sure which way to go to work around that (changing the build type removes the 'builtin' install which helps skip lots of boilerplate). I will ask on luarocks. Thanks for the PR.

teto:
so I've gone the Makefile way and have a working build at https://github.com/teto/nvim-treesitter/tree/release-rockspec-0.8.1. Will try to merge this in nvim-treesitter/test it in nixpkgs

MrcJkb:
@rcarriga plenary.nvim now has an automated release workflow :partying_face: 

https://github.com/nvim-lua/plenary.nvim/pull/439

MrcJkb:
> so I've gone the Makefile way and have a working build at https://github.com/teto/nvim-treesitter/tree/release-rockspec-0.8.1. Will try to merge this in nvim-treesitter/test it in nixpkgs

@teto I'll merge your makefile approach with the automated release workflow and open a PR to nvim-treesitter.

rcarriga:
Awesome to hear! I'll try get some time to integrate an automated release with the semantic release neotest already has. I'd love to have versioned releases :grin:

mrcjkb:
I have just [published a GitHub action](https://github.com/marketplace/actions/luarocks-tag-release) to simplify LuaRocks releases.

mrcjkb:
I have updated this PR with a workflow that automatically publishes to LuaRocks whenever a release is published, along with a LuaRocks badge in the readme (assuming the existence of a `nvim-neotest/neotest` package).

### Notes

* For the release workflow to work, someone with a Luarocks account will have to add their [API key](https://luarocks.org/settings/api-keys) to this repo's [GitHub actions secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository).
* Due to a shortcoming in LuaRocks (https://github.com/luarocks/luarocks-site/issues/188), the `neovim` label has to be added manually, for this plugin to show up in https://luarocks.org/labels/neovim


__Adding the API key (screen shot)__
github-add-luarocks-api-key

rcarriga:
Hey apologies for not looking at this recently, but thanks for keeping it maintained and improving it! I love having it all handled by the action :grin:

rcarriga:
Hmm it doesn't seem to have triggered, any idea why?

mrcjkb:
> Hmm it doesn't seem to have triggered, any idea why?

That's strange. I've been having a similar issue in which my release workflow doesn't publish the action to the marketplace (it just creates a github release).

Maybe the `published` trigger isn't the right one for this workflow. I'll debug it in my fork and open another PR with a fix.

PR #193, opened 2023-01-11
fix: don't error if adapter has no root position

stevearc:
I'm seeing an error when setting `discovery.enabled = false`. I have an adapter that does not have any valid tests discovered, and thus has no root position in the state. This causes the assert here to always fail, which means that I see an error on every `BufAdd` and `BufWritePost`. The repro is as follows:

1. Open a file in a different language from the test adapter
2. Open the summary pane `require("neotest").summary.toggle()`
3. Open any new buffer `:enew`
Observe error

The fix I put in was to silently ignore the adapter if there is no root. I think this is fine (as that is a valid state), but let me know if that is not the case.


rcarriga:
Sorry this totally slipped by me! Yep this fix looks good thanks for the PR :grin:

PR #205, opened 2023-02-10
quickfix: always set QF list after a test run

emmanueltouzery:
otherwise we wouldn't remove obsolete errors after fixing a test.


rcarriga:
Thanks for the fix!

PR #217, opened 2023-03-04
feat: state consumer

rcarriga:


sidlatau:
Thanks for the PR! I tested it, and I am getting results, but I noticed some issues:
 - When tests are running, I do not get intermediate status while tests are running - the status updates to "passed"/"failed" only when all tests finish. But for example, in test summary statuses are being updated in real-time. 
 - When I open the file and immediately run tests, sometimes I do not get `running` test results. 
 - In some files with many files when all tests finish I still see one running test (even when in the test summary there are no running tests).
image

Here is my status function:
```lua
      function()
        local status_ok, neotest = pcall(require, "neotest")
        if not status_ok then
          return ""
        end
        local adapters = neotest.state.adapter_ids()
        if #adapters > 0 then
          local status = neotest.state.status_counts(adapters[1], {
            buffer = vim.api.nvim_buf_get_name(0),
          })
          local sections = {
            {
              sign = "",
              count = status.failed,
              base = "NeotestFailed",
              tag = "test_fail",
            },
            {
              sign = "",
              count = status.running,
              base = "NeotestRunning",
              tag = "test_running",
            },
            {
              sign = "",
              count = status.passed,
              base = "NeotestPassed",
              tag = "test_pass",
            },
          }

          local result = {}
          for _, section in ipairs(sections) do
            if section.count > 0 then
              table.insert(
                result,
                "%#"
                  .. section.base
                  .. "#"
                  .. section.sign
                  .. " "
                  .. section.count
              )
            end
          end

          return table.concat(result, " ")
        end
        return ""
      end,

```

rcarriga:
Thanks for testing @sidlatau! 

> When tests are running, I do not get intermediate status while tests are running - the status updates to "passed"/"failed" only when all tests finish. But for example, in test summary statuses are being updated in real-time.

I've addressed this so you should see updates the same as the summary.

> In some files with many files when all tests finish I still see one running test (even when in the test summary there are no running tests).

This may also be fixed by the previous change, please let me know!

> When I open the file and immediately run tests, sometimes I do not get running test results.

Do you mean that the the statuses go from 0 running/failed/passed to 0 running and x failed/passed? How long do the tests take to run?

rcarriga:
FYI @weilbith I've added a `positions` function to get the tree of positions in a buffer

sidlatau:
@rcarriga I just checked the latest changes, and I am unable to reproduce any of the previous problems, thanks for the update! LGTM for the merge! 🚀

weilbith:
Thank you very much for your work on this! 🙏🏾 
Despite the simple status, I have already 3 different "plugins" planned using this data. 😊

rcarriga:
Great to hear! Would love to see it in use when ready :smile:

PR #223, opened 2023-03-25
fix(luarocks-upload): trigger after release

mrcjkb:
@rcarriga this fixes the issue of the luarocks upload not being triggered by the release (see [my test run](https://github.com/mrcjkb/neotest/actions/runs/4520338868/jobs/7961389493)).

Unfortunately, using `push: tag: ...` as a trigger dosn't work either - probably because the  tag created by the GitHub release isn't "pushed". So I have to manually get the tag and inject it as an environment variable.

I'm marking this as draft, and would advise against merging just yet. It seems like the nvim-treesitter release workflow is still having some issues, which @teto is working on. Until that has been fixed, this workflow will fail.


mrcjkb:
Closing in favour of #307

PR #224, opened 2023-03-25
fix: set window filetype earlier

mawkler:
I had an issue where neotest clashes with the window-resizing plugin [windows.nvim](https://github.com/anuvyklack/windows.nvim/) when opening the the test summary window, despite explicitly telling window.nvim to ignore the `neotest-summary` filetype.

It turns out that neotest sets the filetype `neotest-summary` too late. windows.nvim triggers on `BufWinEnter` and checks the filetype, which wasn't yet set by neotest. I can imagine that more plugins that check the filetype on `BufWinEnter` (or similar events) would have the same issue.

In this PR the window filetype is set as early as possible on open, which fixes the issue. I'm not sure if my solution is the best way to solve it, so feel free to tweak it!

### DIFF
--- a/lua/neotest/lib/window.lua
+++ b/lua/neotest/lib/window.lua
@@ -33,6 +33,13 @@ function PersistentWindow:open()
     return self._win
   end
 
+  if not self._has_opened then
+    if self._bufopts.filetype then
+      async.api.nvim_buf_set_option(self:buffer(), "filetype", self._bufopts.filetype)
+    end
+    self._has_opened = true
+  end
+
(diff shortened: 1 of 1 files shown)


rcarriga:
Thanks for the PR! :smile: Unfortunately setting the filetype happens after opening the window specifically to support users using a filetype autocmd or `ftplugin` to set window settings, otherwise they would have no effect. As that is built-in functionality I'd prefer to support that instead.

What about telling windows.nvim to ignore `nofile` buftypes?

mawkler:
I see, I'll go with the `nofile` solution, thank you!

PR #226, opened 2023-03-28
fix(deprecation): Use query.parse to avoid deprecation warning

crazymerlyn:
See https://github.com/neovim/neovim/pull/22761

Couldn't run the tests. Getting an error about plenary:
```
E492: Not an editor command: PlenaryBustedDirectory tests/ {minimal_init = 'tests/init.vim'}
```


rcarriga:
Thanks for the PR! :smile:

PR #228, opened 2023-04-02
feat: nio async library

rcarriga:
Neotest is migrating away from using plenary's async library, and replacing it with a new library called `nio`. There are several reasons for this change.

## Reasoning

Some major ones are:
- Documentation: There is no vimdoc documentation and minimal documentation in the README.
- Typing: I believe providing type hints and docstrings make a huge difference when using a
  library. Plenary has very few of either.
- Consistency: The API has several inconsistencies that mean I regularly have to
  check source code for how to use it (e.g. some control objects use regular `.`
  syntax such as `mpsc` whereas others use `:` syntax such as `Semaphore`).
- Error handling: Handling errors in async functions requires using a custom `apcall`

This is not to say the plenary library has not been great and I absolutely appreaciate all the effort that went into it but the library is now seemingly no longer actively maintained (e.g. https://github.com/nvim-lua/plenary.nvim/pull/293 and https://github.com/nvim-lua/plenary.nvim/pull/298) and I believe there are significant enough changes to be made to justify packaging a new library.

The new library has a simple goal: providing common asynchronous primitives and asynchronous APIs for Neovim's core. It has been heavily influenced by the previous work for async IO in neovim including plenary and others. There are plans for Neovim to embed an async library within core which will provide a barebones core for libraries to share for compatibility (see https://github.com/neovim/neovim/issues/19624). Nio aims to maintain compatibility with this library.

Features of `nio` include:
- Task objects which track running async functions
- Handling results and errors of tasks along with cancelling.
- Helper functions for coordinating tasks, including gathering results from a list of tasks and getting the result of the first task to complete (and cancelling the others).
- Async versions of many libuv functions, pretty much the same way plenary does it with the added bonus of all of the functions being typed with emmylua annotations so users will have full type inference.
- Creating [a fully documented and typed LSP client interface](https://github.com/nvim-neotest/neotest/blob/92badcc9372b50b11720ec116df84e5cc9d3b8fb/lua/nio/lsp-types.lua#L4) which is generated by a script based off of the LSP spec meta model and then implemented very simply with some metatables.
- Async primitives (futures, queues, events and semaphores) which are often useful in async code, again based off of Python's asyncio modules.

## What does this mean for adapters?

I've been using the library embedded in nvim-dap-ui for a while now but neotest provides a better use case as the async usages are more complex and diverse. Right now I'll be leaving it embedded in neotest, with it being released as a separate plugin in future.

I've checked the source code all the adapters mentioned here (gotten from the neotest README) and most of them won't see any user facing issues as they are using `neotest.async` which has been migrated to point to `nio`. All usage of `fn`, `api` and `uv` will work as before.

Those adapters using `require("plenary.async.tests")` will have broken tests, which can be fixed by switching to `require("nio").tests`.

neotest-jest and neotest-dotnet do use the `control` module which which has changed significantly. I will open PRs to make changes to use the new primitives. 

I'm planning on merging these changes relatively soon so please let me know of any issues but hopefully this will be a painless switch :smile:


lua/neotest/client/strategies/dap/init.lua
```diff
@@ -65,14 +63,10 @@ return function(spec)
       return result_code ~= nil
     end,
     output_stream = function()
-      local sender, receiver = async.control.channel.mpsc()
-      data_accum:subscribe(function(d)
-        sender.send(d)
-      end)
+      local queue = nio.control.queue()
```
scottming, line 66: Wow, this pr  is a huge change. Actually, I never quite understood the async feature in Lua before. The new API looks much easier to understand.

By the way, I don't know if you've seen Elixir's [Task.async](https://hexdocs.pm/elixir/1.14.4/Task.html) module. I've always thought that the process model of Elixir/Erlang is the most understandable concurrency model.

rcarriga: I'm not familiar with Elixir but from a quick look it looks quite similar to Python's `asyncio` and thus this library with the `Task` object already :smile: Is there anything in particular that you think would be worth adapting from Elixir?

scottming: Nope, I'm not familiar with Lua and Python asynchronous. I just said it casually, but the functions they implemented should be similar, although the underlying technology is probably very different.

rcarriga:
neotest-dotnet: @Issafalcon
neotest-deno: @MarkEmmons
neotest-jest: @haydenmeade
neotest-elixir: @jfpedroza
neotest-vitest: @marilari88
neotest-haskell: @mrcjkb
neotest-go: @sergii4 @akinsho
neotest-phpunit: @olimorris
neotest-rspec: @olimorris
neotest-rust: @rouge8
neotest-testthat: @shunsambongi
neotest-dart: @sidlatau
neotest-scala: @stevanmilic
neotest-pest: @theutz

mrcjkb:
Had to replace a `neotest.async` require with `nio`, but otherwise all is good from my side.

olimorris:
Completely agree with the rationale. I've found the documentation for plenary to be subpar for a while now (ChatGPT to the rescue).

I updated my adapters with a conditional for `nio`.

I noticed that my [overseer](https://github.com/stevearc/overseer.nvim) consumer is broken though (cc: @stevearc)

stevearc:
I have a PR up for overseer that uses the new library and generally works, but it spits out this error:
```
[ERROR] Task Neotest dispatch callback on_output: ...vearc/.local/share/nvim/lazy/neotest/lua/nio/control.lua:48: attempt to call local 'callback' (a nil value)
```
I've tried fiddling with the implementation and passing functions around, but I can't seem to find where this is coming from. This highlights my main issue with writing async code in most languages: stack traces lose information and it becomes a lot harder to debug.

@rcarriga Do you have any tips for figuring out where this `control.wait` is getting called from?

stevearc:
Thanks for the help! The overseer PR is ready to go with full support for the new nio library.

Issafalcon:
Hi @rcarriga - Thanks for the update.

I've tested your changes on my end. All looks fine to me. The (few) tests I have still pass also.

rcarriga:
Brilliant, since we're looking good I'll get to merging this

PR #231, opened 2023-04-07
Enhance diagnotic column position

scottming:
If the column is 0, but the code doesn't start on the first column, which isn't what we want. The cursor will be placed to the left of the actual definition.

### before

image

### After this pr

image

### DIFF
--- a/lua/neotest/consumers/diagnostic.lua
+++ b/lua/neotest/consumers/diagnostic.lua
@@ -72,18 +72,6 @@ local function init(client)
   end
 
   function BufferDiagnostics:create_diagnostics(positions, results)
-    local function count_leading_spaces(str, count)
-      count = count or 0
-      local first_char = str:sub(1, 1)
-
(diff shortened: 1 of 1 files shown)


lua/neotest/consumers/diagnostic.lua
```diff
@@ -93,10 +105,11 @@ local function init(client)
               {}
             )
             local mark_code = api.nvim_buf_get_lines(bufnr, mark[1], mark[1] + 1, false)[1]
+
             if mark_code == self.error_code_lines[pos_id][error_i] then
               diagnostics[#diagnostics + 1] = {
                 lnum = mark[1],
-                col = 0,
+                col = count_leading_spaces(mark_code),
```
rcarriga, line 112: Can do `col = mark_code:find("%S") - 1`

rcarriga (changes requested):
Thanks for the PR! I like the idea, just one suggestion :smile:

PR #232, opened 2023-04-11
Fix the symbol range definition format for Elixir

scottming:
The query of Elixir lang is 

```lua
elixir = [[
      ;; query
      (call (identifier) @_func_name
        (arguments (alias) @symbol)
        (#match? @_func_name "^(alias|require|import|use)")
        (#gsub! @symbol ".*%.(.*)" "%1")
      )
    ]],
```

The alias looks like `alias Lexical.SourceFile.Position`, and the symbol is `Lexical.SourceFile.Position`, but what we need is `Position`

So we can't use a more straightforward query to get the actual symbol, we should use #gsub!(I think this is the easest way to archive this.)

By the way, the bug fix for gsub has just been merged: https://github.com/neovim/neovim/pull/23015


rcarriga:
Thanks for the PR! But I think the better solution would be to allow for defining custom functions to parse the tree. I'm worried about other adapters having similar needs that don't quite fit into the same solution as elixir. I've added your code in the latest commit (with credit :smile:) to the branch if you could verify

scottming:
Thank you, then I can continue testing on that branch.

PR #233, opened 2023-04-12
feat: add position to strategy spec

stevearc:
👋 
In the [overseer strategy](https://github.com/stevearc/overseer.nvim/blob/1ddb3671b7198b64c73824a5e5e21f8eb2ab47c3/lua/neotest/client/strategies/overseer.lua#L53), I would really like to be able to include a bit more context about the test being run in the task title. Right now every Neotest run is just called "Neotest", which is...fine, but not ideal. I poked around and I think that the only way to get position information is to pass it in here, but if there's another way I'd be happy to hear it.


rcarriga:
Thanks for the PR :smile: 

I'd rather leave the spec untouched as it's constructed by the adapter. Instead we can pass a second argument to strategies to provide context. This would also open us to allowing users to pass arguments to the strategy directly which could be useful 
```lua
---@class neotest.StrategyContext
---@field position neotest.Position

---@alias neotest.Strategy async fun(spec: neotest.RunSpec, context: neotest.StrategyContext): neotest.Process
```

stevearc:
Refactored!

rcarriga:
LGTM,, thanks again!

PR #236, opened 2023-04-19
docs: add neotest-playwright to list of adapters

thenbe:


rcarriga:
Nice thanks for the PR and the adapter!

PR #243, opened 2023-05-04
Patch 1

nicoandresr:


rcarriga:
Thanks for the PR but I don't want to document adapter specifics in the README. Users can read the documentation of the adapters

nicoandresr:
Ok, but is necessary including some information about adapters could require some extra configuration

PR #245, opened 2023-05-09
Update README.md

nicoandresr:
Readme updating to include important note to set up adapters first


rcarriga:
Thanks for the PR. I've added a smaller note instead and reworded

PR #249, opened 2023-05-15
feat: add output_panel.clear()

marcelbeumer:
Often while re-running tests it's hard to see what is output of the current run and output of the last run.

By manually wiring clear() into my bindings I can choose when to clear and when not.

### DIFF
--- a/lua/neotest/consumers/output_panel/init.lua
+++ b/lua/neotest/consumers/output_panel/init.lua
@@ -6,6 +6,10 @@ local OutputPanel = require("neotest.consumers.output_panel.panel")
 ---@type neotest.OutputPanel
 local panel
 
+---@private
+---@type number
+local chan
+
 local neotest = {}
 
 ---@toc_entry Output Panel Consumer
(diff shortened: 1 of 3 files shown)


lua/neotest/consumers/output_panel/init.lua
```diff
@@ -83,6 +86,16 @@ function neotest.output_panel.toggle()
   end
 end
 
+--- Clears the output panel
+--- >vim
+---   lua require("neotest").output_panel.clear()
+--- <
+function neotest.output_panel.clear()
+  if chan then
+    vim.api.nvim_chan_send(chan, "\x1b[H\x1b[2J")
```
rcarriga, line 95: This only clears the visible section of the output panel. A `clear` function should completely remove the output.
Can do this with
```lua
  nio.api.nvim_buf_set_option(panel.win:buffer(), "modifiable", true)
  nio.api.nvim_buf_set_lines(panel.win:buffer(), 0, -1, false, {})
  nio.api.nvim_buf_set_option(panel.win:buffer(), "modifiable", false)
```

rcarriga (changes requested):
Thanks for the PR, love the idea! Just a small change then LGTM :smile:

thuan1412:
Hi @marcelbeumer, could you continue this PR? If not, how may I assist you to complete it? I really need this feature.

Issue #252, opened 2023-05-24
[IMPROVEMENT] Provide default commands


Zeioth:
It would probably be a good idea to  provide commands for

```lua
require('neotest').run.run(),
require('neotest').run.run(vim.fn.expand '%'),
require('neotest').run.run { strategy = 'dap' },
require('neotest').run.stop(),
require('neotest').run.attach(),
```

This would allow package managers like lazy to load the plugin ONLY when a certain command is executed.

Example:
```lua
{
    "nvim-neotest/neotest",
    cmd = {
       "RunTest" -- For example
    },
}
```

Issue #241, opened 2023-04-25
Tried to use neotest, but having a hard time to keymappings


sittim:
I am having a hard time with figuring out the keybindings?  I have been using the vim-test and my configuration has been like so:

`
nmap <silent> <leader>t :TestNearest<CR>
nmap <silent> <leader>T :TestFile<CR>
nmap <silent> <leader>a :TestSuite<CR>
nmap <silent> <leader>l :TestLast<CR>
nmap <silent> <leader>g :TestVisit<CR>
`

How can I set the NeoTest to trigger to run tests?

fnune:
Here's an example:

```lua
vim.keymap.set("n", "<leader>t", function() require("neotest").run.run() end)
vim.keymap.set("n", "<leader>T", function() require("neotest").run.run(vim.fn.expand('%')) end)
```

Or Vimscript:

```vim
nmap <leader>t :lua require('neotest').run.run()<cr>
nmap <leader>T :lua require('neotest').run.run(expand('%'))<cr>
```

I've tested the Lua but not the Vimscript.

To find more functions you could run, see:

```
:h neotest.run
```

In there, you'll see `run_last` and others.

PR #254, opened 2023-05-26
feat: add command ':Neotest'

MunifTanjim:
Add initial implementation of the `:Neotest` command with completion support.

Instead of adding a lot of command, it's cleaner to have a single one.

The structure will be like this:

`:Neotest [cmd] [subcmd] [...]`

Most of the functionality that I could find are exposed with this command.

---

- Resolve https://github.com/nvim-neotest/neotest/issues/241
- Resolve https://github.com/nvim-neotest/neotest/issues/252
- https://github.com/nvim-neotest/neotest/discussions/227

### DIFF
--- a/plugin/neotest.lua
+++ b/plugin/neotest.lua
@@ -1,0 +1,270 @@
+local cmd_completion_store = {
+  [""] = {
+    "attach",
+    "jump",
+    "output",
+    "output-panel",
+    "run",
+    "stop",
+    "summary",
+  },
+  jump = {
+    "next",
+    "prev",
+  },
+  ["output-panel"] = {
+    "open",
+    "close",
+    "toggle",
+  },
+  run = {
+    "file",
+    "last",
+  },
+  summary = {
+    "close",
+    "mark",
+    "open",
+    "toggle",
+  },
+  ["summary:mark"] = {
+    "clear",
+    "run",
+    "toggle",
+  },
+}
+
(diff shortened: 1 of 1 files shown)


plugin/neotest.lua
```diff
@@ -0,0 +1,270 @@
...
+    "toggle",
+  },
+}
+
+local opt_completion_store = {
+  adapter = function(arg)
+    local part = arg:gsub("^adapter=", "")
+    return vim.tbl_filter(function(cmd)
+      return not not string.find(cmd, "^" .. part)
+    end, require("neotest").run.adapters())
```
rcarriga, line 42: Use `require("neotest").state.adapter_ids()` instead, `neotest.run.adapters` is deprecated

rcarriga (changes requested):
This is brilliant, thank you very much for the PR! :smile: Just one small comment, then LGTM

plugin/neotest.lua
MunifTanjim: Updated

rcarriga:
Thanks again for this, very nice to have :grin:

Zeioth:
@MunifTanjim there is some equivalent to...?

```
require('neotest').run.run({ strategy = 'dap' })
```

MunifTanjim:
@Zeioth that would be `:Neotest run strategy=dap`

PR #262, opened 2023-06-15
fix: args should be table in neotest.run.run. Fixes #261

### LINKED ISSUE #261: [BUG] commit 3649425d28369c290 breaks neotest-plenary
git bisect shows that commit https://github.com/nvim-neotest/neotest/commit/3649425d28369c290ca9ed30e851d9a2bd22de4d introduced an issue with neotest-plenary when trying to run a file.

I get the following neotest warning: `neotest-plenary

<pre>vim/shared.lua:0: after the second argument: expected table, got string</pre>

I've already checked the neotest-plenary code and the error does not seem to originate there.

I'll further look into it, but I wanted to report the bug already so you know.

folke:


rcarriga:
Ah missed this thank you!

PR #270, opened 2023-07-05
refactor: use upstream treesitter api to query lang by ft

amaanq:
nvim-ts master will be deprecated in the future (in favor of main) - though not for a while, it's best to adapt sooner rather than later


wookayin:
probably superseded by #298

PR #279, opened 2023-08-06
Fix bug caused by buf closing

scottming:
After closing the buf, the `mark[1]` becomes `nil`, so we can't just `mark[1] + 1`

image
image


rcarriga:
Thanks for the PR! :smile:

PR #298, opened 2023-10-09
refactor!: remove nvim-treesitter dependency

amaanq:
Nvim-treesitter's 1.0 rewrite will remove all utilities/functionality it provides since vim.treesitter pretty much covers it all now,  you can read more in the *main* branch's readme (default is master)

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -18,27 +18,29 @@ jobs:
 
   tests:
     name: tests
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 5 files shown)


wookayin:
Great PR. One note is that the `vim.treesitter.*` APIs require neovim 0.9 or higher, so this would have an impact on the minimum supported neovim version (so it's technically a breaking change towards v4.0).

amaanq:
Yeah hence why I marked it a breaking change :) I completely forgot I wrote that old PR 3 months ago...

rcarriga:
Thanks for the PR! :smile: The failing test is fixed in master so merging

wookayin:
This PR seems incomplete - we'd need to take care of nvim requirement?

PR #300, opened 2023-10-11
docs(README): link to neotest-java

rcasia:
Hi, I have created an adapter for java and I would like to link it.

I'm not sure how to solve this. Also I'm new to OSS contributions.
A couple of weeks ago, I found there was another adapter for java, which it's not quite working.

I have been slowly developing neotest-java in my free time because I intended to use it in my day to day workflow and I have been testing it on my projects.

So much thanks for neotest! :heart:

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ See the adapter's documentation for their specific setup instructions.
 | scala           |    [neotest-scala](https://github.com/stevanmilic/neotest-scala)     |
 | haskell         |    [neotest-haskell](https://github.com/mrcjkb/neotest-haskell)      |
 | deno            |    [neotest-deno](https://github.com/MarkEmmons/neotest-deno)        |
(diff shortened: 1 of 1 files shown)


wookayin:
What's the relationship with https://github.com/andy-bell101/neotest-java?

rcasia:
While both projects share the same name, they have been developed independently and in parallel, without knowledge of each other's existence.

At the very least, I initiated my project earlier, and I was unaware that someone else started working on something similar at a later time.

rcarriga:
Since this project seems to cover the existing one's feature set and extend it, I'll merge this. I'd like to avoid having two adapters for the same use case as it'll likely just lead to duplicated effort. 

@andy-bell101 Please let me know if you disagree, I'm not at all familiar with Java :smile:

andy-bell101:
Sure thing. I haven't used Java as much as planned to originally so I kind of forgot about my plugin

PR #306, opened 2023-10-25
status: only place marks for tests with their own range

### LINKED ISSUE #305: status: with child tests present, markers always indicate the last result of the latest-defined child
I ran into this in my neotest-python PR (https://github.com/nvim-neotest/neotest-python/pull/36#issuecomment-1779664683):

> There is still something funky going on with status markers: only the latest defined of the child tests which have run updates the marker. So if you run the entire set of children, the last result will be the only one reflected: if you then run any child but the last one, the status marker does not even update to reflect that the test is running. If you run individual children (from the summary window), then the status marker is updated iff you haven't run any later-listed children.

I think I've figured out what's going on here. My change in this area in https://github.com/nvim-neotest/neotest/commit/6676edc3078b8936df989df639900dc9a70dfc8a#diff-866d3516ad8bfe1d181d7b54bddf2bb25543d9e18828cd9876bab6401707dfc4 wasn't correct: it means that we set a mark on the test's first line for the parent _and_ every child that has a result in `results`, which explains the behaviour I've observed. Instead, we should simply skip setting status markers for child tests, deferring to the parent.

OddBloke:
This fixes #305, by ensuring that rangeless tests will not have markers written at all, avoiding overwriting their parent's mark.

This does introduce another, less significant, issue: the marker for a test will always indicate the result of the last time that the parent test was explicitly executed, including being absent if the parent test has never been specifically requested.  (Currently, this likely only occurs when using the summary window, so users can reference that for more precise result info.  This could become a more annoying issue with, for example, a "run all failed tests" command in future which might bypass parent test execution.)

### DIFF
--- a/lua/neotest/consumers/status.lua
+++ b/lua/neotest/consumers/status.lua
@@ -15,7 +15,7 @@ local function init(client)
 
   local namespace = nio.api.nvim_create_namespace(sign_group)
 
-  local function place_sign(buf, pos, range, adapter_id, results)
+  local function place_sign(buf, pos, adapter_id, results)
     local status
     if results[pos.id] then
       local result = results[pos.id]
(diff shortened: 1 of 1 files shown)


rcarriga:
> This does introduce another, less significant, issue: the marker for a test will always indicate the result of the last time that the parent test was explicitly executed, including being absent if the parent test has never been specifically requested. (Currently, this likely only occurs when using the summary window, so users can reference that for more precise result info. This could become a more annoying issue with, for example, a "run all failed tests" command in future which might bypass parent test execution.)

I think this is fine for now. I'd be curious how other editors solve this issue but it's a pretty small issue as you said

PR #307, opened 2023-10-28
ci: fix luarocks-release workflow

mrcjkb:
This fixes the luarocks release workflow:

- Removes treesitter from the dependencies
- Makes sure the workflow can be triggered by GitHub releases

### DIFF
--- a/.github/workflows/luarocks-release.yaml
+++ b/.github/workflows/luarocks-release.yaml
@@ -1,17 +1,29 @@
+---
 on:
   release:
-    types: [published]
+    types:
+      - created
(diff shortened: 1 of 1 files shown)


mrcjkb:
Not sure why the workflow just failed. It seems unrelated - maybe a flake?

mrcjkb:
@rcarriga any chance of getting this merged?

We're working on [a luarocks-base plugin manager](https://github.com/nvim-neorocks/rocks.nvim), and it would be nice to be able to install neotest with it :smile:

PR #308, opened 2023-10-31
feat(watcher): add symbol queries for ruby rspec tests

wassimk:
I've been using these symbols queries via ruby-lsp[^1] on a Ruby on Rails RSpec application for the past few weeks with success. The watcher feature is so useful!

Most RSpec classes look like this:

```ruby
RSpec.describe MyClass do
  it "my first test" do
    # ...
  end
end
```

It also covers name spaced modules like so, `MyClass` is the returned symbol.

```ruby
RSpec.describe NameSpaceModule1::NameSpace2::MyClass do
  it "my first test" do
    # ...
  end
end
```

References:
[^1]: https://github.com/Shopify/ruby-lsp

### DIFF
--- a/lua/neotest/config/init.lua
+++ b/lua/neotest/config/init.lua
@@ -313,6 +313,22 @@ local default_config = {
         end
         return symbols
       end,
+      ruby = [[
+        ;query
+        ;rspec - class name
+        (call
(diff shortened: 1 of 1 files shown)


rcarriga:
Thanks for the PR! :smile:

PR #313, opened 2023-11-12
Add watcher support for Go

jan-xyz:
This should capture most of the types, functions, vars and constants that can be affected by change. I tested it on my own corpus for a while and am super happy with the result! Thank you for this awesome feature!!


rcarriga:
Brilliant thanks for the the PR!

PR #315, opened 2023-11-13
fix: nvim-dap hijacks summary window when using the debug mapping

luismeyer95:
neotest-hijack-bug

## Bug

Whenever the debug mapping is used inside the summary window and a breakpoint is hit, nvim-dap hijacks the summary window to focus the breakpoint.

## Repro

- Set config option `switchbuf = 'uselast'` in nvim-dap (default)
- Open test file
- Set a breakpoint in test
- Open + move to summary window
- Use debug mapping `d` on test

## Expected

The window containing the test buffer should be used to focus the breakpoint.

## Fix

`nvim-dap` has the `switchbuf` config option to to control how a breakpoint is jumped to, the default is `uselast` and it probably makes the most sense as a default for neotest too. It works like this: when `dap.run()` is executed, if the current buffer is a normal buffer or `dap.run()` is not provided with a different `filetype` than the current buffer, then the current window is used to focus the breakpoint, otherwise the previous (`#`) window is used. Initially `filetype` was not provided, so the current window was picked.

The change broke some tests due to async related issues. I'm not familiar with async lua so I didn't dive too deep, I saw some sleep hacks being thrown around in tests so I replicated that and it fixed them, lmk if that should be handled differently.

### DIFF
--- a/lua/neotest/client/runner.lua
+++ b/lua/neotest/client/runner.lua
@@ -116,6 +116,7 @@ function TestRunner:_run_spec(spec, tree, args, adapter_id, adapter, results_cal
   local position = tree:data()
   local context = {
     position = position,
+    adapter = adapter,
   }
 
   local proc_key = self:_create_process_key(adapter_id, position.id)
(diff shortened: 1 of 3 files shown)


rcarriga:
Thanks for the PR!

Instead of passing the filetype in the context, let's pass the adapter. The filetype logic is a weird, niche requirement only for the dap strategy so I don't think it should live in the `TestRunner` object. The filetype retrieval logic can then live in the dap strategy, and the change to the tests can be reverted :smile:

luismeyer95:
Good call, done! Thanks for the awesome plugin 👍🏻

PR #316, opened 2023-11-14
Output close mapping

ashton:
Hey there, when accessed via `Summary`, output pop up get focused and the only way of quitting it is to change to another window/buffer. I found that odd and kind of annoying, so I'm adding a configurable mapping (disabled by default, so we don't have any changes on the current behavior) for closing the output pop up when it is the current focused window / buffer.

### DIFF
--- a/lua/neotest/config/init.lua
+++ b/lua/neotest/config/init.lua
@@ -105,6 +105,7 @@ define_highlights()
 ---@class neotest.Config.output
 ---@field enabled boolean
 ---@field open_on_run string|boolean Open nearest test result after running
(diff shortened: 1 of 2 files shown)


rcarriga:
Thanks for the PR but I'd rather not add another configuration option just for this. I have this in my config 
```lua
  vim.api.nvim_create_autocmd("FileType", {
    pattern = "neotest-*",
    callback = function(opts)
      vim.keymap.set("n", "q", function() vim.api.nvim_win_close(0, true) end,
      { buffer = opts.buf, })
    end,
  })

```
and it covers the attach and summary windows too

PR #325, opened 2023-12-01
Update README.md to include treesitter dependency

Wayfarer98:
Tests might not be found if treesitter is not added as a direct dependency to neotest. Most distros fixed this themselves, but the documentation in neotest would be an improvement.


rcarriga:
This wasn't documented because it is not a dependency if adapters don't require it, however I believe most adapters do use treesitter so I understand the need :smile:

rcarriga:
No you can use neotest without treesitter, however most adapters use treesitter parsers and the majority of users will be using nvim-treesitter to install them.

PR #331, opened 2023-12-10
doc(readme): add neotest-gtest

alfaix:
Adding [neotest-gtest](https://github.com/alfaix/neotest-gtest) to the list as it's now more-or-less functional.

Also wanted to mention - I wrote integration tests while working on the plugin, they allow writing tests like [this](https://github.com/alfaix/neotest-gtest/blob/main/tests/integration/plugin_spec.lua#L27) from files like [this](https://github.com/alfaix/neotest-gtest/blob/main/tests/integration/cpp/src/test_one.cpp#L13). Would it make sense to contribute the testing tools to this plugin? Of course, `configure_executable` and things like that don't make sense for other adapters, but I think the general "write test project" -> "ensure it's running" flow would help with testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,29 +87,30 @@ See the adapter's documentation for their specific setup instructions.
 ### Supported Runners
 
 | Test Runner     |                               Adapter                                |
-| :-------------- | :------------------------------------------------------------------: |
-| pytest          |   [neotest-python](https://github.com/nvim-neotest/neotest-python)   |
-| python-unittest |   [neotest-python](https://github.com/nvim-neotest/neotest-python)   |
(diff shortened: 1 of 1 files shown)


rcarriga:
Thanks for the PR! :smile: 

> Would it make sense to contribute the testing tools to this plugin? 

I'm not sure they'd make much sense to upstream. Having a quick look, they're quite complex and I'd be worried about how abstract they could be that would make sense for other adapters without having to hack low level custom logic into them.

Although they are pretty cool, I personally wouldn't have much interest in them for the adapters that I maintain as I use them all and so I just test them through usage. They however aren't the most complex adapters I've seen so maybe not the best example.

IMO the ones you've written are better as example for adapter authors that may have an interest rather than the basis of a rather complex "library" of test functions

Issue #195, opened 2023-01-19
Persist neotest test results in neovim session


Willem-J-an:
When I run tests, and save my neovim session, then close and re-open the session, the tests that previously succeeded show a question mark again in the summary panel. Is it possible to persist the result of the tests in a session, so they would remain as passed when I re-open a session?

If not, could it be nice as feature request?

Thanks!

andryuha-salo:
I think that's not really feasible, considering anything can happen to the code between nvim launches (git pull), so cached results may be outdated

Willem-J-an:
There is nothing preventing me from doing a git pull without closing neovim. Test results being outdated is not necessarily a problem, just something to keep in mind, paired with an option to reset results when needed.

PR #341, opened 2024-01-03
feat: add result cache consumer

Willem-J-an:
Attempt to solve #195.

Creating as draft, because the client interface does not seem to expose any way to update results. Any pointers on how to solve this? I have some ideas but not super happy with them:
1. The client exposes a method to update_results; might expose too much internals to consumers
2. The client exposes a method to load results from cache (as done now); bit weird because it's logic only my consumer needs, but arguably exposes less internals.

There is also currently still a bug that output_panel consumer is unable to read output from closed tmp files present in the cached results. Will take a look at it after some input on the questions above.


Willem-J-an:
@rcarriga any feedback on this? Thanks!

lua/neotest/consumers/result_cache.lua
```diff
@@ -0,0 +1,88 @@
...
+    os.remove(file)
+  end
+end
+
+neotest.result_cache.clear = nio.create(neotest.result_cache.clear, 1)
+
+--- Loads previously cached test results
+---@async
+---@return nil
+function neotest.result_cache:fetch()
```
rcarriga, line 74: Same question here, think it'd better to just expose a config variable and we automatically fetch the results once the client starts

rcarriga (changes requested):
Thanks for the PR! Overall love the idea, just some minor fixes/changes. I also think this would make sense as part of the state consumer and in the state consumer config we can have something like `persist_results: boolean`, rather than a whole separate consumer for it. It can still be a separate module to keep it tidy

lua/neotest/consumers/result_cache.lua
Willem-J-an: I've been using the plugin for a while now; I haven't really used the clear feature much. A functional use case I experienced is like: 

You're working on a specific feature, you run all the tests, a couple of them fail, you want to track which ones failed, so you can work on fixing them, even after restarting the editor. I decide to cache the test results. I didn't have autoload, so I could choose if I wanted to save/fetch my cache to continue that effort. If I get side-tracked on another task, I could decide not to load my cache, run tests, not save results. Restart editor, load my old cache and get back to the initial task.

From this use case I would say: keep the manual option to fetch.
I don't mind to add a configuration to auto fetch on start-up, but I think I would not use it myself.

If auto fetch is enabled, it makes more sense to have a clear cache feature, but rm -rf ~/.cache/nvim/neotest could be sufficient as well.

rcarriga: Thanks for the explainer, makes sense to me! With that in mind, I do think it makes sense to keep the `clear` function as well

Willem-J-an: I have kept the clear function in!

Willem-J-an:
> Thanks for the PR! Overall love the idea, just some minor fixes/changes. I also think this would make sense as part of the state consumer and in the state consumer config we can have something like `persist_results: boolean`, rather than a whole separate consumer for it. It can still be a separate module to keep it tidy

I have added a config to auto load cache on startup; not sure now what functionality you would expect from a persist_results boolean. It should automatically cache all results as soon as they come in; rather than having to execute a command to cache it?

PR #358, opened 2024-02-02
feat(watch): allow passing in `run_predicate` that takes a bufnr

mrcjkb:
This is useful e.g. for checking if the buffer has any error diagnostics from the LSP client, in which case one would probably not want to rerun the tests yet.


rcarriga:
Nice thanks for the PR :smile:

Issue #366, opened 2024-02-19
nio and busted?


HiPhish:
Continuing the discussion from #349. I am currently writing an adapter for [busted](https://lunarmodules.github.io/busted/) and I want to write tests for it in busted as well. I can run tests in general without issue following [this](https://hiphish.github.io/blog/2024/01/29/testing-neovim-plugins-with-busted/) blog post. The issue is when executing asynchronous functions like `neotest.lib.treesitter.parse_positions`.

Here is a simple test file:

```lua
local nio = require 'nio'

describe('Discovery of test positions', function()
	local tempfile

	before_each(function()
		-- Create temporary file
		tempfile = vim.fn.tempname()
	end)

	after_each(function()
		-- Delete temporary file
		if vim.fn.filereadable(tempfile) ~= 0 then
			vim.fn.delete(tempfile)
		end
	end)

	it('Always succeeds', function()
		assert.is_true(vim.endswith('abc', 'c'))
	end)
end)
```

As you can see this busted test is able to call into Neovim's `vim` module. Other than that it is a regular busted test. Now let's adding an actial test.

## A naive attempt

```lua
	it('Discovers nothing in an empty file', function()
		local result = adapter.discover_positions(tempfile)
		print(vim.inspect(result))
	end)
```

This fails because `discover_positions` calls `lib.treesitter.parse_positions` without asynchronous context. The return values of `lib.treesitter.parse_positions` are wrong and the file descriptor will be returned as the error. Issue #349 describes the same behaviour.

## Using `nio.tests.it`

```lua
	nio.tests.it('Discovers nothing in an empty file', function()
		vim.fn.writefile({''}, tempfile, 's')
		local result = adapter.discover_positions(tempfile)
		print(vim.inspect(result))
	end)
```

This throws another error about `it` being `nil` inside to body of `nio.tests.it`.

```
Error → ./test/unit/discover_positions_spec.lua @ 4
Discovery of test positions
...e/nvim/site/pack/testing/start/neotest/lua/nio/tests.lua:35: attempt to call global 'it' (a nil value)
```

## Replicate `nio.tests.it`

If the global `it` only exists within the test, how about we replicate the entire function inside the test?

```lua
	it('Does something async', function()
		local success, err
		local task = nio.tasks.run(function()
				assert.is_true(true)
				assert.equal('x', 'y')
				return 'Yay!'
			end,
			function(success_, err_)
				success = success_
				if not success_ then
					err = err_
				end
			end)
[...]

PR #369, opened 2024-02-20
Dynamically bind busted hook functions

HiPhish:
Follow-up to #366, I have picked the second solution because it maintains the existing interface. I have successfully tested this change with busted, but not with Plenary busted because that one does not work for me in general.


rcarriga:
Nice this looks great! Thanks for the PR :smile:

PR #374, opened 2024-03-01
Add async context function to tests module

HiPhish:
As a user of the nio library when writing tests I might want to run an asynchronous function within a synchronous function and wait for the result.  This function acts as a synchronous wrapper around the asynchronous function.

This is either an alternative to #369 or a complement. It lets me use the normal busted `it` function and write code like this:

```lua
it('Does something, function()
	local tree = nio.tests.with_async_context(adapter.discover_positions, tempfile)
	local result = tree:to_list()
	assert.are.same(expected, result)
end)
```

Here the function `adapter.discover_positions` is asynchronous, so we wrap it in `nio.tests.with_async_context` to provide it with an asynchronous context. The resulting call is synchronous and can be used in tests. We can pass arguments to the wrapped function and we return its result to the caller.

Why not use `nio.tests.it` instad? `nio.tests.it` runs the entire tests in an asynchronous context, which messes up error reporting, at least when used with the real busted. And it turns all assertion failures into errors when they should be test failures instead.


rcarriga:
LGTM thanks again for the PR!

PR #389, opened 2024-04-16
fix: check buffer valid before set diagnostic

towry:
The error happens when deleting the buffer.


rcarriga:
Thanks for the PR :+1:

PR #390, opened 2024-04-16
feat(config): setup adapters with function

olexsmir:
Hey :wave:

I have a problem with setting up the plugin with lazy.nvim's opts because it would load adapters at the same time they weren't loaded, and because of that, I would fail. It might be prevented by wrapping adapters setup in function, so I made it.

The solution looks like this:
```lua
{
  "nvim-neotest/neotest",
  opts = {
    -- now `adapters` can be set up to {} as previously or to a function that will return {} of adapters
    adapters = function()
      return {
        require "neotest-plenary",
        require "neotest-go",
      }
    end,
  }
}
```

Also I believe it might help set up adapters on different conditions

### DIFF
--- a/doc/neotest.txt
+++ b/doc/neotest.txt
@@ -219,7 +219,7 @@ neotest.config                                                  *neotest.config*
 
                                                             *neotest.CoreConfig*
 Fields~
-{adapters} `(neotest.Adapter[])`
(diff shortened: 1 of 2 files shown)


rcarriga:
Thanks for the PR, but this seems like more of an issue with your local setup and dependency resolution, I don't want to have this plugin adapt to the quirks of package managers unless there's a very good reason

PR #394, opened 2024-04-20
Link NeotestDir highlight to built-in Directory

nulty:
## Context

Directories and files should be differentiated by default in the UI


rcarriga:
Thanks for the PR, but the color scheme can be changed in your personal config.

PR #404, opened 2024-05-11
fix(autocommand): use `afile` equivalent from callback data

przepompownia:


rcarriga:
Ah thanks for the PR! But merged #416 first :sweat_smile:

przepompownia:
The point is that it's merged anyway in any way.

PR #405, opened 2024-05-17
safer place_sign

devenv:
I'm occasionally getting an `Invalid 'line': out of range`, probably because buffer changed and the line is not there.
Also if I don't open the test in a buffer - it errors despite the existing check (any help understanding why is appreciated).


rcarriga:
Thanks for the PR!

PR #427, opened 2024-06-27
fix: properly handle nil value from cpu_info()

massix:
On Android devices, the function `vim.loop.cpu_info()` returns a nil value, since it is unable to read from `/proc` for security reasons. We should handle that failure properly in Neotest and avoid panicking.


rcarriga:
Thanks for the PR!

PR #433, opened 2024-07-07
chore(window): replace deprecated usage nvim_buf_set_option

codymikol:
this is now deprecated in 0.10 and can be replaced with nvim_set_option_value

per https://neovim.io/doc/user/deprecated.html#deprecated-0.10

Fixes: N/A


rcarriga:
Thank you :+1:

PR #434, opened 2024-07-08
fix(window): handle when a PersistedWindow's buffer has been deleted

codymikol:
Adds an additional check to the self.buffer function of PersistedWindow to see if the current bufnr still exists. This prevents a crash in cases where the persisted window's buffer has been deleted by an external script.

Fixes N/A


rcarriga:
Thanks for the PR!

PR #438, opened 2024-07-15
Add optional env variables to spawn

asmodeus812:
Adding a env variables option to spawn

Related/Blocking: https://github.com/rcasia/neotest-java/pull/124

### DIFF
--- a/lua/neotest/lib/process/init.lua
+++ b/lua/neotest/lib/process/init.lua
@@ -1,6 +1,8 @@
 local nio = require("nio")
 
+--- Base module definition
 local neotest = { lib = {} }
+
 ---@toc_entry Library: Processes
 ---@text
 --- Utilities to run external processes easily.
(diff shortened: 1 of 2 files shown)


asmodeus812:
@rcarriga could you take a look at this small change, it is blocking a bigger one in neotest-java

asmodeus812:
@rcarriga  any chance this gets looked at ?

lua/neotest/lib/process/init.lua
```diff
@@ -1,16 +1,51 @@
...
 ---@text
 --- Utilities to run external processes easily.
 --- More complex use cases should use the lower level jobstart or vim.loop.spawn
 ---@class neotest.lib.process
 neotest.lib.process = {}
 
+--- @return table<string,string>
+local function base_env()
+  --- @type table<string,string>
+  local env = vim.fn.environ()
```
rcarriga, line 16: Use `vim.uv.os_environ()` instead to avoid using a vimscript function (means we don't need to call the scheduler)

lua/neotest/lib/process/init.lua
```diff
@@ -1,16 +1,51 @@
...
+  env["NVIM"] = vim.v.servername
```
rcarriga, line 17: Could you explain the need for these?

PR #448, opened 2024-08-26
README: add neotest-bazel

sluongng:
I have started a new adapter to work with Bazel specifically.

Also adjusted the column alignment of kotlin


sluongng:
cc: @rcarriga

rcarriga:
Thanks for the PR and adapter!

PR #450, opened 2024-09-04
fix(query): explicitly opt-in to legacy behavior

dpetka2001:
After https://github.com/neovim/neovim/commit/6913c5e1d975a11262d08b3339d50b579e6b6bb8 in Neovim nightly, neotest returns `No tests found` (tested with `neotest-python`). Since it's mentioned in the PR that the option `all = false` will be removed in the future, maybe a more robust approach would be preferable, but I don't have the knowledge to implement it. This is just a hotfix until someone else comes with something better.

### DIFF
--- a/lua/neotest/config/init.lua
+++ b/lua/neotest/config/init.lua
@@ -333,7 +333,7 @@ local default_config = {
           ]]
         )
         local symbols = {}
-        for _, match, metadata in query:iter_matches(root, content) do
+        for _, match, metadata in query:iter_matches(root, content, _, _, { all = false }) do
           for id, node in pairs(match) do
             local name = query.captures[id]
 
--- a/lua/neotest/lib/treesitter/init.lua
+++ b/lua/neotest/lib/treesitter/init.lua
@@ -56,7 +56,7 @@ local function collect(file_path, query, source, root, opts)
       range = { root:range() },
     },
   }
-  for _, match in query:iter_matches(root, source) do
+  for _, match in query:iter_matches(root, source, _, _, { all = false }) do
     local captured_nodes = {}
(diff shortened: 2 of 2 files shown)


GitMurf:
@dpetka2001 does this relate to either of these treesitter commits?

- https://github.com/nvim-treesitter/nvim-treesitter/commit/3826d0c42ac635f560479b5b6ab522f6627a3466
- https://github.com/nvim-treesitter/nvim-treesitter/commit/fd9663acca289598086b7c5a366be8b2fa5d7960

dpetka2001:
No, these are just adjustments in `nvim-treesitter` plugin as well. The original behavior stems from Neovim builtin treesitter and it's this [specific commit](https://github.com/neovim/neovim/pull/30193).

GitMurf:
So is this a "breaking change" for potentially many plugins that use treesitter? or is this something that was recently changed with neotest as well that has created a "bug"? I am just trying to figure out what the intended impact is overall and if I need to watch out for issues like this with other plugins that utilize treesitter in particular ways? Thanks!

dpetka2001:
Yes, it's a breaking change. It's also mentioned in the PR itself 
> BREAKING CHANGE: Any plugin which uses Query:iter_matches() must update their call sites to expect an array of nodes in the match table, rather than a single node.

And you can also see it from the title of the convetional commit `feat(treesitter)!`, where the exclamation mark is used to identify breaking changes. Another plugin that I noticed had problems with this breaking change from the ones that I use was `vim-matchup`. I'm sure there might be other as well, but don't use them so don't really know.

rcarriga:
Thanks for the PR!

PR #451, opened 2024-09-06
feat: use client.supports_method instead of client.server_capabilities

rcasia:
Fixes #425.

For reference: https://github.com/neovim/neovim/commit/ddd92a70d2aab5247895e89abaaa79c62ba7dbb4.

PR #456, opened 2024-09-23
Fix: Test watch won't start on v0.10.1

straightchlorine:
On the version v0.10.1, supports_method is still a table, rather than a function. Trying to access the table field as a function resulted in following error:

```
attempt to call field 'supports_method' (a table value)
```

This behavior is now fixed.

### DIFF
--- a/lua/neotest/consumers/watch/init.lua
+++ b/lua/neotest/consumers/watch/init.lua
@@ -32,10 +32,13 @@ neotest.watch = {}
 local function get_valid_client(bufnr)
   local clients = nio.lsp.get_clients({ bufnr = bufnr })
   for _, client in ipairs(clients) do
-    local has_definition_support = client.supports_method
-        and client.supports_method("textDocument/definition")
-      -- for compatibility with Neovim versions earlier than v0.10.1
-      or client.server_capabilities.definitionProvider
+    local has_definition_support
+    if client.server_capabilities then
+      -- for compatibility with Neovim versions earlier and equal to v0.10.1
+      has_definition_support = client.server_capabilities.definitionProvider
+    elseif type(client.supports_method) == "function" then
+      has_definition_support = client.supports_method("textDocument/definition")
+    end
 
     if has_definition_support then
       logger.debug("Found client", client.name, "for watch")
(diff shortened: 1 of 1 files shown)


rcasia:
My fault. I was distracted in #451 because of neotest not working with nightly version. I was preparing a fix for this as well and I have just one correction.

lua/neotest/consumers/watch/init.lua
```diff
@@ -32,14 +32,21 @@ neotest.watch = {}
...
   for _, client in ipairs(clients) do
-    local has_definition_support = client.supports_method
-        and client.supports_method("textDocument/definition")
-      -- for compatibility with Neovim versions earlier than v0.10.1
-      or client.server_capabilities.definitionProvider
+    local has_definition_support
+    if client.server_capabilities then
+      -- for compatibility with Neovim versions earlier and equal to v0.10.1
+      has_definition_support = client.server_capabilities.definitionProvider
+    elseif type(client.supports_method) == "function" then
```
rcasia, line 39: 1. Actually this should be using client.supports_method.textDocument_definition as a function. Like [here](https://github.com/nvim-neotest/nvim-nio/blob/a428f309119086dc78dd4b19306d2d67be884eee/tests/lsp_spec.lua#L97-L113).

2. Actually type(client.supports_method.textDocument_definition) will always be a `function`, so we should check nvim version instead.

3. Actually supports_method was already included in nvim v0.10.1 . See https://github.com/neovim/neovim/commit/ddd92a70d2aab5247895e89abaaa79c62ba7dbb4 .
    
Something like this:
```lua
    local has_definition_support
    if nio.fn.has("nvim-0.10.1") == 0 then
      -- for compatibility with Neovim versions earlier than v0.10.1
      has_definition_support = client.server_capabilities ~= nil and client.server_capabilities.definitionProvider
    else
      has_definition_support = client.supports_method.textDocument_definition({ bufnr = bufnr })
    end



```

straightchlorine: Yeah, you're absolutely right. Just tested it on 0.10.1, just in case, fixes the problem. Looks better as well, I think.

tftiwnbe (approved):
Reviewed the changes, the issue is resolved. Everything works correctly

rcasia (approved):

rcarriga:
Thanks for the PR!

PR #458, opened 2024-09-25
fix(query): use `nil` to skip optional parameters when calling the iter_matches method

archilkarchava:
https://github.com/nvim-neotest/neotest/pull/450 sometimes breaks test parsing behavior after changing directories, because we're passing `_` instead of a `nil` value to the `iter_matches` function.
This PR fixes the problem.


rcarriga:
Thank you :smile:

PR #465, opened 2024-11-01
fix(ui): combine virtual text bg highlight with cursor line bg

89iuv:
Before:
Screenshot 2024-11-01 at 13 42 48

After:
Screenshot 2024-11-01 at 13 42 30


rcarriga:
Thanks for the PR!

PR #477, opened 2024-11-28
fix(output_panel): recreate window and term channel if panel closed manually

YaroSpace:
A fix for 2 issues that appear if `output_panel` is closed by the user manually and not with `output_pannel.toggle/close()`:
1.  The buffer id for output panel becomes invalid, so there is an error when trying to open the panel again.  Currently the check with `vim.fn.bufexists()` does not catch buffers that exist, but are unloaded.  Fix - recreate the buffer.
2.  The channel id for term bound to panel buffer becomes invalid, so there is an error when trying to write to that channel.  Fix - recreate the term.

Also, a fix with associated issue: if you reaload the plugin with say `Lazy reload neotest`, the old panel buffer still exists, so when a new buffer is created with the same name, there is an error of trying to rename existing buffer.  Fix - delete buffer with the same name, when ceating a new one.

### DIFF
--- a/lua/neotest/consumers/output_panel/init.lua
+++ b/lua/neotest/consumers/output_panel/init.lua
@@ -25,7 +25,12 @@ local init = function(client)
     if partial then
       return
     end
-    if not chan then
+
+    local channel_is_valid = function(chan_id)
+      return chan_id and pcall(vim.api.nvim_chan_send, chan_id, "\n")
+    end
+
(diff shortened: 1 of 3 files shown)


rcarriga:
Awesome thanks again for the PR!

YaroSpace:
It is a pleasure! Thank you for the great plugin!

PR #479, opened 2024-11-30
fix(watcher): race condition between positions update and watch run on BufWritePost

YaroSpace:
When watching a spec and editing it at the same time, threre are 2 BufWritePost events triggered on file save. One to update NeotestClientState:update_positions and the other for the watcher, which runs the spec.
If the watcher spec is run before the positions are updated, it is run with the stale tree and there is a mismatch in positions ids.

To fix this: I added a wait for an discover_positions event.


YaroSpace:
It is a pleasure. Thank you for accepting!

PR #481, opened 2024-12-31
feat: triggering QuickFixCmdPost event after setting quickfix list

brunotvs:
This update introduces the triggering of the QuickFixCmdPost event after the quickfix list is set.

The primary purpose of this change is to ensure tools like Trouble can update properly, especially in scenarios where there are no failing tests remaining.


rcarriga:
Great thanks for the PR!

PR #488, opened 2025-03-01
Fix issue when multiple test processes output data to neotest.

tlxpro:
I was experimenting with https://github.com/HiPhish/neotest-busted and I had issues with data being corrupted when read after multiple commands to busted completed concurrently. Seems like the issue is with multiple calls to uv_fs_write. According to https://github.com/libuv/libuv/issues/3820#issuecomment-1311563948, the responsabilty to control concurrent access to same file must lie with the user.

NVIM v0.11.0-nightly+1fb606b 
Build type: Release
LuaJIT 2.1.1713773202
OSX

### DIFF
--- a/lua/neotest/client/strategies/integrated/init.lua
+++ b/lua/neotest/client/strategies/integrated/init.lua
@@ -29,8 +29,9 @@ return function(spec)
   assert(not open_err, open_err)
 
   output_accum:subscribe(function(data)
-    vim.loop.fs_write(output_fd, data, nil, function(write_err)
-      assert(not write_err, write_err)
+    nio.run(function()
+      local ok, write_err = vim.uv.fs_write(output_fd, data, nil)
+      assert(ok, write_err)
     end)
   end)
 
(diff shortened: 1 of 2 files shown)


rcarriga:
I'm not sure if this actually fixes the issue as you're just sending the write to a new coroutine, you could still have several concurrent writes. Your test is failing for me locally.

I've just pushed a change to prevent concurrent writes, can you check with the latest master if you still the issue?

rcarriga:
Ah sorry yes, I've changed it now so it will await the write. I still don't get the test passing with the sync version you have above or with the `nio` version I've implemented unfortunately

tlxpro:
It might be due to how my system open files, I have a not so recent version of OSX. 10.15.

It didn't change anything for me when I tried this, but this might work for you.

```lua
local open_err, output_fd = nio.uv.fs_open(output_path, "w", tonumber("4010666", 8)) -- with O_SYNC
assert(not open_err, open_err)
```

tlxpro:
Just tested again. Is `nio.uv.fs_write` equivalent to `vim.uv.fs_write` because the former doesn't work.

PR #493, opened 2025-03-21
fix(watch): update Haskell query

mrcjkb:
tree-sitter-haskell has undergone a rewrite:

- https://github.com/tree-sitter/tree-sitter-haskell/pull/120

This PR updates the `watch` query for Haskell.


rcarriga:
Nice one thank you!

Issue #482, opened 2025-01-05
[BUG] watch fails to watch dependency files in large go packages


asfaltboy:
**NeoVim Version**

```
NVIM v0.10.3
Build type: RelWithDebInfo
LuaJIT 2.1.1731601260
Run "nvim -V1 -v" for more info
```

**Describe the bug**

On larger go projects, the watch functionality is very slow and borderline unusable, as a trigger for re-running tests in larger packages. See more details in context section below.

**To Reproduce**

Unfortunately, I struggle to get 'watch' feature working _with dependency files_ using a minimal config (it only watches for changes in the actual test file), though the dependency files are watched when I use the LazyVim distro config ([here](https://github.com/LazyVim/LazyVim/blob/d0c366e4d861b848bdc710696d5311dca2c6d540/lua/lazyvim/plugins/extras/lang/go.lua#L56)). I think it has something to do with my LSP client config, would appreciate a minimal config example for supporting watch (it doesn't seem to work out of the box without LSP client/server configuration).

Here's my minimal config:
```lua
vim.opt.runtimepath:remove(vim.fn.expand("~/.config/nvim"))
vim.opt.packpath:remove(vim.fn.expand("~/.local/share/nvim/site"))

local lazypath = "/tmp/lazy/lazy.nvim"

if not (vim.uv or vim.loop).fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    "--branch=stable", -- latest stable release
    lazypath,
  })
end

vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
  "nvim-neotest/neotest",
  dependencies = {
    "nvim-neotest/nvim-nio",
    "nvim-lua/plenary.nvim",
    "antoinemadec/FixCursorHold.nvim",
    "nvim-treesitter/nvim-treesitter",
    "neovim/nvim-lspconfig",
    -- Install adapters here
    "fredrikaverpil/neotest-golang",
  },
  config = function()
    -- Install any required parsers
    require("nvim-treesitter.configs").setup({
      ensure_installed = { "go" },
    })

    require("neotest").setup({
      -- Add adapters to the list
      adapters = {
        require("neotest-golang"),
      },
      log_level = "debug",
    })

    require("lspconfig").gopls.setup({
      staticcheck = true,
      directoryFilters = { "-.git" },
    })
  end,
})
```

[...]

PR #496, opened 2025-04-04
fix: watcher unnecessarily waiting for discover position event to be set

brunotvs:
#479 introduced a small bug where the watcher would wait for the ```discover_positions_event``` by the listener set up by the init function. This PR fixes it by checking if the changed file is relevant to the discover_position_event.


brunotvs:
It should fix #482

rcarriga:
Thanks for the PR!

PR #501, opened 2025-04-26
Summary: follow on open

### LINKED ISSUE #271: [BUG] Summary is not expanded when opened after the tests are ran
**NeoVim Version**
0.9.1

**Describe the bug**
With the `expand_errors = true`, when the summary is opened while the tests are running, the failed test are automatically expanded.
However, if the summary is opened after the test runner has finished, then the summary is folded.

**To Reproduce**
Make sure to have `pytest` somewhere, and `minimal.lua` with `neotest-python`, then:
```bash
mkdir tests
touch tests/__init__.py
echo "def test_foobar(): assert False" > tests/test_foobar.py
nvim --clean -u minimal.lua tests/test_foobar.py
```
1. `:Neotest`
2. `:NeotestSummary`

**Expected behavior**
I would have expected the summary to be expanded.

What do you think?

benjiwolff:
Fixes #271

### DIFF
--- a/lua/neotest/consumers/summary/init.lua
+++ b/lua/neotest/consumers/summary/init.lua
@@ -52,7 +52,6 @@ local function init(client)
       if summary.win:is_open() then
         func()
       else
-        vim.api.nvim_clear_autocmds({ group = group })
         vim.api.nvim_create_autocmd("User", {
           pattern = "NeotestSummaryOpen",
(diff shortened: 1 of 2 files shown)


azmeuk (approved):

rcarriga:
Thanks for the PR!

PR #508, opened 2025-05-17
docs(README): add neotest-vstest

Nsidorenco:
Adds an alternative adapter for dotnet which uses the vstest runner


rcarriga:
Thank you!

Issue #91, opened 2022-08-17
[BUG] Running colorscheme after running setup removes highlight information.


yuki-yano:
**NeoVim Version**
$ nvim --version
NVIM v0.8.0-dev-2317-g5e98a4187
Build type: Release
LuaJIT 2.1.0-beta3

**Describe the bug**

After setup in init.vim, setting colorschema will remove highlight
I think it is correct to execute the following code in the ColorSchema event as well
https://github.com/nvim-neotest/neotest/blob/d018868c1abe79bf730c9acbbcdaa58ebec8d1f4/lua/neotest/config/init.lua#L3-L20

**To Reproduce**

```vim
call plug#begin('~/.vim/plugged')
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-treesitter/nvim-treesitter'
Plug 'antoinemadec/FixCursorHold.nvim'
Plug 'nvim-neotest/neotest'
Plug 'haydenmeade/neotest-jest'
call plug#end()

lua << EOF
require("neotest").setup({
  adapters = {
    require("neotest-jest")
  }
})
EOF
```

- Open jest test file
- Run `lua require('neotest').summary.open()`.
- Run `colorscheme blue`.
- neotest's highlight is removed

**Expected behavior**

Highlight is reconfigured when running colorscheme.

rcarriga:
This should be working as expected in master :smile:

PR #512, opened 2025-06-06
enhance highlight group config

letieu:
I see now, neotest just force user to use only default color, and colorscheme have no effect. 

So now I do some change to:
- make default color is link with builtin neovim highlight group, like `DiagnosticError`, `DiagnosticWarn`, ... So user can use whatever colorscheme, it still look good and fit with colorscheme.
- move `define_highlights` call to `setup` function, so it only load call when seup.
- remove `nvim_create_autocmd("ColorScheme" ...)` , so colorscheme can custom color if they want.

You can conpare in the images, left is old, and right is after I update
image
image

### DIFF
--- a/lua/neotest/config/init.lua
+++ b/lua/neotest/config/init.lua
@@ -1,26 +1,98 @@
 ---@tag neotest.config
 ---@toc_entry Configuration Options
 
+local default_highlights = {
+  {
+    name = "NeotestPassed",
+    link = "DiagnosticOk",
+    desc = "Passed icon",
+  },
+  {
+    name = "NeotestFailed",
+    link = "DiagnosticError",
+    desc = "Failed icon",
+  },
+  {
+    name = "NeotestRunning",
+    link = "DiagnosticWarn",
+    desc = "Running test icon",
+  },
+  {
+    name = "NeotestSkipped",
+    link = "Comment",
+    desc = "Skipped test icon",
+  },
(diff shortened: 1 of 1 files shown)


rcarriga:
Thanks for the PR. I have a couple of requests:

> I see now, neotest just force user to use only default color, and colorscheme have no effect. 

This is not true, it is using the `default` argument.
> The [default] argument is used for setting the default highlighting for a
group.	If highlighting has already been specified for the group the command
will be ignored.  Also when there is an existing link.

This is likely already an issue with your setup. Are you defining them also with the `default` argument and your calls are loaded after neotest? Moving the call to `setup` suggests this. Please restore it to the original location.

> move define_highlights call to setup function, so it only load call when seup.
remove nvim_create_autocmd("ColorScheme" ...) , so colorscheme can custom color if they want.

This was a purposeful feature so users can switch colorschemes and neotest will apply the defaults again in case they're not defined by the colorscheme. See #91 for context. Please restore this functionality.

letieu:
I just use this config

```

require("neotest").setup({
  adapters = {
    require("neotest-python")({
      dap = { justMyCode = false },
    }),
    require("neotest-plenary"),
    require("neotest-vim-test")({
      ignore_file_types = { "python", "vim", "lua" },
    }),
  },
})

```

And my color look like the image above with default neovim color.

PR #514, opened 2025-06-25
fix: reset chan when caling output_panel.clear()

89iuv:
I found 2 issues related to the output_panel which are hard to explain but i will give it my best.

Issues
1. Sometimes the output_pannel would get cleared in between executions, sometimes it would just append new results.
2. After manually clearing the output_pannel, next executions would have their output truncated.

Investigation
1. This is really random, my only guess is that sometimes the channel is recreated which also clears the buffer and sometimes it is not. Why this is happening I have no idea.
2. It seems that after the buffer has been cleared, the terminal channel linked to it will not worked correctly anymore.

Solution
The solution for the second issue seems to be: to not reuse the terminal channel when clearing the buffer. 
By using this solution where the channel is declared globally (same as panel) i looks like the first issue is also fixed.

Disclaimer
I am not a lua or neovim expert so this is my effort to brute force a solution for this issue.
I am using lazyvim so this may be related to how lazyvim loads the plugin. Also if lazy loading is non deterministic this could explain the first issue, but again this are just assumptions.

### DIFF
--- a/lua/neotest/consumers/output_panel/init.lua
+++ b/lua/neotest/consumers/output_panel/init.lua
@@ -7,6 +7,7 @@ local OutputPanel = require("neotest.consumers.output_panel.panel")
 local panel
 
 local neotest = {}
+local chan
 
 ---@toc_entry Output Panel Consumer
 ---@text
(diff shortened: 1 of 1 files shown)


rcarriga:
Apologies for the delay but thanks for the PR, LGTM :smile:

Issue #515, opened 2025-06-26
Pipeline error


Dynge:
Pipeline has failed on nightly nvim release for a couple of months. This seems to hinder the release to luarocks.

Would be awesome to have this fixed as I am using Rocks.nvim to install from luarocks.

Thank you

PR #518, opened 2025-06-30
build: use correct targets for neovim builds.

codymikol:
the artifacts used to have the name linux64, but were split into x86_64 and arm64. This should make the nightly builds run successfully.

Fixes: #515


rcarriga:
Amazing thank you!

PR #519, opened 2025-07-04
feat: build_position provide treesitter node metadata

### LINKED ISSUE #516: [FEATURE] Use treesitter metadata.text if present
Hello, I'm working on [neotest-kotlin](https://github.com/codymikol/neotest-kotlin) and have a rather challenging predicament.

For complete context,

```kotlin
class KotestWordSpec : WordSpec({
    "first namespace" should {
        "pass" {
            1 shouldBe 1
        }
    }

    "second namespace" `when` {
        "nested namespace" should {
            "pass" {
                1 shouldBe 1
            }

            "fail" {
                1 shouldBe 2
            }
        }
    }

    "third namespace" `when` {
        "nested namespace" should {
            "pass" {
                1 shouldBe 1
            }

            "fail" {
                1 shouldBe 2
            }
        }
    }
})
```

This is a [Kotest WordSpec](https://kotest.io/docs/framework/testing-styles.html#word-spec) the challenge here underneath the hood is that for this specific spec type Kotest changes the name of the namespaces

"first namespace" -> "first namespace should"

They append on the word that's used to declare the test (I think we're going to hit similar issues with BehaviorSpec).

My solution here is to leverage Treesitter directives specifically at

https://github.com/nvim-neotest/neotest/blob/1d4b3bd89afa8bfa12fffd2bb1ccd26ac3c92ce5/lua/neotest/lib/treesitter/init.lua#L25C50-L39

Where you all reference `vim.treesitter.get_node_text` if you all can get the metadata and if that exists and `metadata.text` exists use that rather than the node_text.

This would allow for other plugins (assuming they need to do this at all) to apply modifications to any captured groups.

Specifically, I'm thinking of doing for the "should" case

```scm
(infix_expression
  (string_literal) @namespace.name (#gsub! @namespace.name "$" " should")
  (simple_identifier) @function_name (#eq? @function_name "should")
  (lambda_literal)
) @namespace.definition
```

Here using the "#gsub!" directive to append on " should" which would match the output of Gradle

```text
org.example.KotestWordSpec > first namespace should > pass PASSED
```

that would allow us to parse this easily! I think this is a very flexible solution that would allow for other plugin hook-ins to Neotest to do other minor modifications if necessary. I'd be happy to implement this if this seems reasonable! 😃

NickHackman:
Closes #516 

My specific use case is to allow for appending/prepending text to `@test.name` and `@namespace.name` for [neotest-kotlin](https://github.com/codymikol/neotest-kotlin) where the testing frameworks we're trying to support append text to namespace and test names. See the issue for a complete example!

This is a non-breaking change, as it passes a new arg to `build_position` of TSMetadata which in my personal use case I'll only be using `text` from.

Here's my expected implementation using this new feature

```lua
local function get_match_type(captured_nodes)
  if captured_nodes["test.name"] then
    return "test"
  end
  if captured_nodes["namespace.name"] then
    return "namespace"
  end
end

local function build_position(file_path, source, captured_nodes, metadata)
  local match_type = get_match_type(captured_nodes)
  if match_type then
    ---@type string
    local name = metadata[match_type .. ".name"].text or vim.treesitter.get_node_text(captured_nodes[match_type .. ".name"], source)
    local definition = captured_nodes[match_type .. ".definition"]

    return {
      type = match_type,
      path = file_path,
      name = name,
      range = { definition:range() },
    }
  end
end
```

## Changes

- Provides Treesitter metadata to `build_position`
- Tests for multiple Treesitter directives

### DIFF
--- a/lua/neotest/lib/treesitter/init.lua
+++ b/lua/neotest/lib/treesitter/init.lua
@@ -94,7 +94,7 @@ end
 
 ---@class neotest.lib.treesitter.ParseOptions : neotest.lib.positions.ParseOptions
 ---@field fast? boolean Use faster parsing (Should be unchanged unless injections are needed)
----@field build_position? fun(file_path: string, source: string, captured_nodes: table<string, userdata>, metadata: vim.treesitter.query.TSMetadata): neotest.Position|neotest.Position[]|nil Builds one or more positions from the captured nodes from a query match.
+---@field build_position? fun(file_path: string, source: string, captured_nodes: table<string, userdata>, metadata: table<string, vim.treesitter.query.TSMetadata>): neotest.Position|neotest.Position[]|nil Builds one or more positions from the captured nodes from a query match.
 
 --- Build a parsed Query object from a string
 ---@param lang string
--- a/lua/neotest/lib/treesitter/init.lua
+++ b/lua/neotest/lib/treesitter/init.lua
@@ -22,7 +22,7 @@ local function get_match_type(captured_nodes)
   end
 end
 
-local function build_position(file_path, source, captured_nodes)
+local function build_position(file_path, source, captured_nodes, metadata)
   local match_type = get_match_type(captured_nodes)
   if match_type then
     ---@type string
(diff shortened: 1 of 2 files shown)


lua/neotest/lib/treesitter/init.lua
```diff
@@ -92,7 +94,7 @@ end
 
 ---@class neotest.lib.treesitter.ParseOptions : neotest.lib.positions.ParseOptions
 ---@field fast? boolean Use faster parsing (Should be unchanged unless injections are needed)
----@field build_position? fun(file_path: string, source: string, captured_nodes: table<string, userdata>): neotest.Position|neotest.Position[]|nil Builds one or more positions from the captured nodes from a query match.
+---@field build_position? fun(file_path: string, source: string, captured_nodes: table<string, userdata>, metadata: vim.treesitter.query.TSMetadata): neotest.Position|neotest.Position[]|nil Builds one or more positions from the captured nodes from a query match.
```
NickHackman, line 97: Would love feedback on naming here! I think `metadata` may be too vague? Maybe it should be `node_metadata` or `nodes_metadata`?

codymikol: since its the name of the type, possibly treesitter_metadata?

PR #520, opened 2025-07-10
fix(consumers/run): check spec.strategy before run

winter-loo:
This patch is to check the type of `spec.strategy`  before the async runners. Before this patch, when an adapter provider does not return the correct type of `spec.stragegy`, neotest will raise the exception below:

image

NOTE the error **after the second argument: expected table, got string**.  That error comes from the [code](https://github.com/nvim-neotest/neotest/blob/61d1951131e6770c815243d84a27ccd0436ec61c/lua/neotest/client/runner.lua#L109):

```lua
      vim.tbl_extend("force", spec.strategy or {}, config.strategies[args.strategy] or {})
```

After this patch, the user will see a better error tip:

image

related commit: https://github.com/mrcjkb/rustaceanvim/pull/803

After the above rustaceanvim fix, neotest gets:

image

This new error tip results in the second commit.

And finally, we get:

image

### DIFF
--- a/lua/neotest/client/strategies/init.lua
+++ b/lua/neotest/client/strategies/init.lua
@@ -39,10 +39,7 @@ function ProcessTracker:run(pos_id, spec, args, stream_processor, context)
   self._process_semaphore.with(function()
     instance = strategy(spec, context)
     if not instance then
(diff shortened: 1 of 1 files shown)


winter-loo:
@OddBloke could this PR be reviewed

PR #533, opened 2025-08-14
feat: Support overriding error severity

rlch:
This allows error severity to be modified by runners. Particularly useful for runners like `neotest-golang` which would otherwise produce error diagnostics when the test succeeds:

image


rlch:
@rcarriga Please let me know if you'd like any changes. This is a pretty simple PR

Issue #531, opened 2025-08-10
[BUG] commit ce17830 causes issues


fredrikaverpil:
## NeoVim Version

```
NVIM v0.11.3
Build type: Release
LuaJIT 2.1.1741730670
Run "nvim -V1 -v" for more info
```

### Tests not found

Prior to the commit, tests are found perfectly fine using the [neotest-golang](https://github.com/fredrikaverpil/neotest-golang) adapter. But with commit ce17830, no Go tests are found.

This is pretty severe and have forced me to pin neotest to the previous commit; 52fca67.

### Test suite visibility in summary panel

In [neotest-golang](https://github.com/fredrikaverpil/neotest-golang) I have both plenary tests and go tests. Prior to this commit I was able to view both these test suites in Neotest's summary panel:

<img width="399" height="138" alt="Image" src="https://github.com/user-attachments/assets/26610a5b-d9ee-4367-af27-710d878d6e97" />

But with commit ce17830, I only see one (and you can see the test count is 0):

<img width="428" height="176" alt="Image" src="https://github.com/user-attachments/assets/7b3cebb1-4aab-424f-8cec-7af5820b1572" />

## Additional info

I'm using the `main` branch of nvim-treesitter, not the `master` branch.

## Repro

```sh
nvim --clean -u minimal.lua
```

You can edit the following example file to include your adapters and other required setup.
```lua
vim.opt.runtimepath:remove(vim.fn.expand("~/.config/nvim"))
vim.opt.packpath:remove(vim.fn.expand("~/.local/share/nvim/site"))

local lazypath = "/tmp/lazy/lazy.nvim"

if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    "--branch=stable", -- latest stable release
    lazypath,
  })
end

vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
  {
    "nvim-neotest/neotest",
    commit = "ce178308ea30f96667f1c11f022880d8c04cafd3",  -- problematic commit
  },
  dependencies = {
    "nvim-lua/plenary.nvim",
    "nvim-treesitter/nvim-treesitter",
    -- Install adapters here
    "nvim-neotest/neotest-plenary",
    "fredrikaverpil/neotest-golang",
  },
  config = function()
    -- Install any required parsers
    require("nvim-treesitter.configs").setup({
[...]

PR #536, opened 2025-08-20
fix(subprocess): load adapters

rcarriga:
See #531

### DIFF
--- a/lua/neotest/client/init.lua
+++ b/lua/neotest/client/init.lua
@@ -535,35 +535,45 @@ function neotest.Client:_update_adapters(dir)
       return i, entry.root
     end, adapters_with_root))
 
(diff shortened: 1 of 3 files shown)


neckhair:
👍 
Seems to work for me. Using the RSpec adapter.

asharkhan3101:
still doesn't work with `neotest-go` adapter.

rcarriga:
Thanks for the stacktrace @fredrikaverpil, could you try again with the latest commit? I think it's because of some lazy loading added by nvim-treesitter

fredrikaverpil:
@rcarriga the latest commit works as expected now with both `master` and `main` branch of nvim-treesitter 🎉

PR #548, opened 2025-10-04
fix: remove private member access from fast_parse()

ribru17:
This function was written at a time when `:parse()` always parsed all injections, which was slow. Now, by default, children are not parsed. Children are only parsed if a range is given, or if `true` is passed.

We should thus use the builtin functionality for `fast_parse()`, since right now it accesses several private members of the language tree, one of which has been removed already `_valid` and others like `_trees` will be removed in the future. Removing them will reduce breakages as nightly improves parsing performance.

Fixes https://github.com/neovim/neovim/issues/36015


tan-wei:
Great if merged.

PR #603, opened 2026-04-21
style(sign): add test sign when no results for a test

Dynge:
This small PR adds a sign to the gutter of a test tube when no results. This marks nicely which methods are recognized as tests and are available to be run.


rcarriga:
Again thanks for the PR!

PR #604, opened 2026-04-21
style(summary): add icons for dirs, files and namespaces

Dynge:
Instead of using an unknown icon in the summary use the icon for the position type.

This makes it clearer what the positions represent and looks nicer (imo).


rcarriga:
LGTM thanks for the PR!

PR #577, opened 2026-01-20
Fix(subprocess): remove tree-sitter dependency

archie-judd:
# Description

Neovim 0.11+ bundles tree sitter parsers in the runtime, and no longer requires nvim-treesitter for basic parsing. This PR updates the subprocess initialisation process to handle both cases.

### Changes

- No longer search for, or add treesitter to the subprocess rtp.
- Discover parser directories via `nvim_get_runtime_file("parser/*", true)` and add their parents to rtp.
- Extract `resolve_plugin_root()` as a separate function for clarity.
- Rename `add_to_rtp` to `add_paths_to_rtp` with clearer signature
- Add regression test for parsing in subprocess.

# Context

The subprocess needs parsers available for treesitter parsing of test files. Previously this relied on nvim-treesitter being installed. Now it works with:
- Neovim 0.11+ bundled parsers (no nvim-treesitter required)
- nvim-treesitter managed parsers (backwards compatibility)
- Both together

Fixes:

https://github.com/nvim-neotest/neotest/issues/531

PR #606, opened 2026-05-08
fix(subprocess): fully remove tree-sitter dependency

adridevlyon:
## Description

In https://github.com/nvim-neotest/neotest/pull/577 there was an attempt to remove tree-sitter dependency that's not used by neotest itself

Problem is : there is still some code requiring it and it causes a [crash](https://github.com/nvim-neotest/neotest/issues/597#issuecomment-4305857980) (at least using Neovim 0.12.1 with Lazy) because
- `nvim-treesitter` is not found neither in the sub-folders nor in the global lua installation
- no root path is provided to resolve this module since [#577 ](https://github.com/nvim-neotest/neotest/pull/577)

### Changes

- remove the code that's related to `nvim-treesitter` and crashes
- adapt the config in the readme

## Remaining question

@rcarriga I can see that there is still a mention of `nvim-treesitter` in the ["Parsing Tests"](https://github.com/nvim-neotest/neotest/blob/ad991822b7076b1d940b33a9d6d0d30416d5df81/README.md?plain=1#L297) part of the readme, especially for a code example. To be fair, I have no idea how this example should be adapted, would you mind help me on this before merging this PR?

Fixes : https://github.com/nvim-neotest/neotest/issues/597


rcarriga:
Thanks for the PR and apologies for the delay!

PR #608, opened 2026-05-11
docs: fix optional total_range LuaCATS annotation

mrcjkb:
This field appears to be optional (if not, it's a breaking change).
My linter started complaining about it...


rcarriga:
Correct, thanks for the PR :smile:

Issue #563, opened 2025-11-10
[BUG] Meet several problems on Windows


Xiaomony:
**NeoVim Version**
NVIM v0.11.5
Build type: Release
LuaJIT 2.1.1741730670
Run "nvim -V1 -v" for more info

**Describe the bug**
1. When I 

**To Reproduce**
Please provide a minimal `init.lua` to reproduce which can be run as the following:
```sh
nvim --clean -u minimal.lua
```

You can edit the following example file to include your adapters and other required setup.
```lua
vim.opt.runtimepath:remove(vim.fn.expand("~/.config/nvim"))
vim.opt.packpath:remove(vim.fn.expand("~/.local/share/nvim/site"))

local lazypath = "/tmp/lazy/lazy.nvim"

if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    "--branch=stable", -- latest stable release
    lazypath,
  })
end

vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
  "nvim-neotest/neotest",
  dependencies = {
    "nvim-lua/plenary.nvim",
    "nvim-neotest/nvim-nio",
    "nvim-treesitter/nvim-treesitter",
    -- Install adapters here
  },
  config = function()
    -- Install any required parsers
    require("nvim-treesitter.configs").setup({
      ensure_installed = {},
    })

    require("neotest").setup({
      -- Add adapters to the list
      adapters = {},
    })
  end,
})
```

Please provide example test files to reproduce. 

**Expected behavior**

**Logs**
1. Wipe the `neotest.log` file in `stdpath("log")` or `stdpath("data")`.
2. Set `log_level = vim.log.levels.DEBUG` in your neotest setup config.
3. Reproduce the issue.
4. Provide the new logs.

**Additional context**

Issue #268, opened 2023-07-04
[BUG] no such file or directory, when running test on Windows


janislaus:
Great plugin! 
I'm using neotest together with neotest-python. Everything works fine on Ubuntu, but not on Windows.

**NeoVim Version**
`NVIM v0.9.1`
**Describe the bug**
Trying to run a test on Windows results in the following log message in neotest.log:
```
ERROR | 2023-07-04T17:07:38Z+0200 | ...te\pack\packer\start\neotest/lua/neotest/client/init.lua:311 | Couldn't find positions in path C:\Users\myUserName\projects\foo\tests\test_neotest.py ...ck\packer\start\neotest/lua/neotest/lib/process/init.lua:41: ENOENT: no such file or directory
stack traceback:
	[C]: in function 'error'
	...ck\packer\start\neotest/lua/neotest/lib/process/init.lua:41: in function 'run'
	...\packer\start\neotest-python/lua/neotest-python/base.lua:17: in function 'module_exists'
	...\packer\start\neotest-python/lua/neotest-python/init.lua:63: in function 'get_runner'
	...\packer\start\neotest-python/lua/neotest-python/init.lua:114: in function 'discover_positions'
	...te\pack\packer\start\neotest/lua/neotest/client/init.lua:302: in function <...te\pack\packer\start\neotest/lua/neotest/client/init.lua:266>
	[C]: in function 'xpcall'
	...te\pack\packer\start\neotest/lua/neotest/client/init.lua:266: in function '_update_positions'
	...te\pack\packer\start\neotest/lua/neotest/client/init.lua:319: in function <...te\pack\packer\start\neotest/lua/neotest/client/init.lua:317>
```

It says that the file does not exist, but it definitely does. This is the test I'm trying to run:
![image](https://github.com/nvim-neotest/neotest/assets/34862469/5f28e8f6-4e40-4174-adc4-8fa479900890)
The project foo is located at C:\Users\myUserName\projects

This is my neotest configuration:
```
require("neotest").setup({
    adapters = {
        require("neotest-python")
    },
    benchmark = {
        enabled = true
    },
    consumers = {},
    default_strategy = "integrated",
    diagnostic = {
        enabled = true,
        severity = 1
    },
    discovery = {
        concurrent = 0,
        enabled = true
    },
    floating = {
        border = "rounded",
        max_height = 0.6,
        max_width = 0.6,
        options = {}
    },
    highlights = {
        adapter_name = "NeotestAdapterName",
[...]

PR #613, opened 2026-06-16
fix(windows): resolve drive casing mismatch, watch consumer failures and subprocess crashes

SparkyCloudy:
> Transparency note: I'm a Windows user who ran into these issues while using neotest with neotest-java. I don't have deep knowledge of the neotest codebase — I used an AI coding agent (Google Antigravity) to help diagnose and fix these problems. The fixes work on my setup, but I'd appreciate maintainers reviewing whether the approach is sound and if there are edge cases I may have missed. Happy to adjust anything based on feedback.

Below is a brief explanation of the problem and the proposed fix.

## Problem

On Windows, neotest has path normalization issues that cause position lookups and watch mode failures:

### 1. Drive letter casing mismatch breaks path lookups

Neovim APIs return paths with inconsistent drive letter casing. For example, `vim.fn.expand("%:p")` may return `e:\projects\...` while `vim.fn.getcwd()` returns `E:\projects\...`. Since Lua's string comparison is case-sensitive, `vim.startswith()` checks and tree key lookups fail silently — neotest can't find positions, can't match buffers to test trees, and reports "No tests found" even though the files exist.

**Affected code paths:**
- `Client:get_position()` — position ID lookup fails
- `Client:_update_positions()` — position discovery fails
- `path.real()` — file existence checks fail

### 2. Watch consumer can't find LSP clients

`get_lsp_client()` in `watch/init.lua` compares buffer paths from `nvim_buf_get_name()` against tree keys. On Windows, the drive letter casing mismatch causes `tree:get_key(path)` to always return `nil`, so the watcher reports "No valid LSP client found" and watch mode silently breaks.

The same issue occurs in `watcher.lua` where `_get_linked_files()` and the `BufWritePost` autocmd handler compare buffer paths against dependency URIs and tree data paths.

### ~~3. Subprocess crashes when `add_paths_to_rtp` receives function objects~~ *(Reverted per review)*

~~When adapters add entries to the subprocess runtime path via `add_paths_to_rtp()`, some adapter paths are actually function objects (e.g., `is_test_file` overrides) rather than path strings. The original code unconditionally concatenated them with `rtp = rtp .. "," .. path`, causing a Lua type error crash in the subprocess.~~

~~Additionally, the nvim-treesitter plugin root was not being added to the subprocess `rtp` during initialization, which could cause Treesitter parser resolution to fail in the child process.~~

---

### Centralized path normalization

Added `neotest.lib.files.path.normalize(path)` in `lib/file/init.lua`:
- Uppercases the drive letter (`e:` → `E:`)  
- Converts forward slashes to backslashes (`/` → `\`)
- Only active on `win32`, no-op on other platforms
- Guards against non-string input

Applied `normalize()` at every boundary where paths enter from Neovim APIs:
- `Client:get_position()` — normalizes `position_id` before lookup
- `Client:_update_positions()` — normalizes `path` before discovery
- `get_lsp_client()` — normalizes buffer path before tree key lookup
- `Watcher:_get_linked_files()` — normalizes path and URI-derived paths
- `Watcher:watch()` BufWritePost handler — normalizes buffer path and tree data path before comparison
- `path.real()` — normalizes the result of `fnamemodify`

### ~~Subprocess rtp safety~~ *(Reverted per review)*

~~Fixed `add_paths_to_rtp()` in `lib/subprocess.lua`:~~
- ~~If a path entry is a `function`, attempt to resolve it to a plugin root directory via `resolve_plugin_root()`~~
- ~~If resolution fails, skip the entry instead of crashing~~
- ~~Added `nvim-treesitter` setup function to `plugin_funcs` so the Treesitter plugin root is included in the subprocess rtp~~

---

## Files Changed

| File | Change |
|------|--------|
| `lua/neotest/lib/file/init.lua` | Added `normalize()` to `path` utilities; applied in `path.real()` |
| `lua/neotest/client/init.lua` | Normalize paths in `get_position()` and `_update_positions()` |
| `lua/neotest/consumers/watch/init.lua` | Normalize buffer path in `get_lsp_client()` |
| `lua/neotest/consumers/watch/watcher.lua` | Normalize paths in `_get_linked_files()` and BufWritePost handler |
| `tests/unit/lib/files/init_spec.lua` | Added unit tests for `neotest.lib.files.path.normalize()` |
| ~~`lua/neotest/lib/subprocess.lua`~~ | ~~Handle function entries in `add_paths_to_rtp()`; add nvim-treesitter to subprocess rtp~~ *(Reverted)* |

---

## Testing

- Added unit tests in `tests/unit/lib/files/init_spec.lua` testing drive casing and path separator normalization.
- Tested manually on Windows 11 and WSL2 with:
  - Neovim `0.12`, LazyVim
  - `neotest-java` adapter with a Maven project
  - Verified test discovery, position lookups, and watch mode all work with mixed-case drive letter paths

---

## Related Issues

- ~~Partially addresses #268 — `[BUG] no such file or directory, when running test on Windows`~~ *(Reverted)*
- Related to [#563](https://github.com/nvim-neotest/neotest/issues/563) — `[BUG] No tests found on Windows` (resolves drive letter casing mismatch during path lookups)

---

## Disclosure

Co-authored with AI coding assistant (Google Antigravity).

### DIFF
--- a/lua/neotest/client/init.lua
+++ b/lua/neotest/client/init.lua
@@ -188,6 +188,7 @@ end
 function neotest.Client:get_position(position_id, args)
   self:_ensure_started()
   args = args or {}
+  position_id = lib.files.path.normalize(position_id)
   if position_id and vim.endswith(position_id, lib.files.sep) then
     position_id = string.sub(position_id, 1, #position_id - #lib.files.sep)
   end
@@ -243,6 +244,7 @@ end
 ---@param path string
 ---@private
 function neotest.Client:_update_positions(path, args)
+  path = lib.files.path.normalize(path)
   if not lib.files.exists(path) then
     return
   end
--- a/lua/neotest/consumers/watch/init.lua
+++ b/lua/neotest/consumers/watch/init.lua
@@ -66,6 +66,7 @@ end
 local function get_lsp_client(tree)
   for _, buf in ipairs(nio.api.nvim_list_bufs()) do
     local path = nio.fn.fnamemodify(nio.api.nvim_buf_get_name(buf), ":p")
+    path = lib.files.path.normalize(path)
     if tree:get_key(path) then
       local client = get_valid_client(buf)
       if client then
--- a/lua/neotest/consumers/watch/watcher.lua
+++ b/lua/neotest/consumers/watch/watcher.lua
@@ -3,6 +3,8 @@ local logger = require("neotest.logging")
 local nio = require("nio")
 local config = require("neotest.config")
 
+
+
 ---@class neotest.consumers.watch.Watcher
 ---@field lsp_client nio.lsp.Client
 ---@field autocmd_id? string
@@ -71,9 +73,9 @@ function Watcher:_get_linked_files(path, root_path, args)
       end
     end
   end
-  local paths = { path }
+  local paths = { lib.files.path.normalize(path) }
   for uri in pairs(dependency_uris) do
-    local p = vim.uri_to_fname(uri)
+    local p = lib.files.path.normalize(vim.uri_to_fname(uri))
     if uri ~= path_uri and args.filter_path(p, root_path) then
       paths[#paths + 1] = p
     end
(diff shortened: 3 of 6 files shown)


rcarriga:
The path handling changes look good :+1: 

Please revert the changes to the subprocess module.
The treesitter issue has been resolved by another PR, and adapters passing incorrect values to the `add_paths_to_rtp` is an issue with the adapter, not neotest

james-pettit:
Just chiming in that I have tested this, and it fixed the issues I was having on neotest for Windows.

PR #615, opened 2026-06-25
docs(README): adds odin adapter

joseildofilho:
Hi Guys,
i'm linking an adapter to odin


rcarriga:
Great thanks for the PR!

PR #617, opened 2026-07-02
docs(readme): add neotest-nix adapter

khaneliman:
List neotest-nix with supported Nix runners so users can find the adapter from the main neotest README.


rcarriga:
Great thanks!

PR #620, opened 2026-07-03
ci: restrict workflow.yaml to auto-run only on nvim-neotest

khaneliman:
Pushing branches to a personal fork was triggering these jobs (and, in release/luarocks-upload's case, failing them outright since forks don't carry the release/publish secrets). Gate push/pull_request-triggered runs behind github.repository so they only fire automatically on the main org repo, and add workflow_dispatch so a fork owner can still run the pipeline manually when they want to.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -4,9 +4,11 @@ on:
     branches:
       - master
   pull_request: ~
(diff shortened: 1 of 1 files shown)


rcarriga:
Thanks for all of the PRs! :smile:

PR #622, opened 2026-07-13
Prevent buffer processing if that buffer represents the contents of a directory

Melandel:
- for instance, using a filesystem plugin like vim-dirvish

This change prevents untimely error message notification coming from the execution of `self:get_nearest` when the user enters a buffer that represents a directory content instead of a file content

image


rcarriga:
LGTM thanks for the PR!

